PR #34, opened 2017-04-12
[hotfix] Repair app.js load script

corysimmons:
When I minified imagery I changed all `assets` paths to `dist` paths. Forgot to check console for bugs. `app.js` wasn't getting loaded. This fixes it.


siddharthlatest:
awesome, thanks!

PR #78, opened 2017-08-18
Fix defaultSelected and URLParams behavior in NestedList

divyanshu013:


divyanshu013:
@metagrover using deep comparisson now, please check again

metagrover:
Looks good. Merging!

PR #88, opened 2017-09-11
Add native events support for input fields

dhruvdutt:
[#210](https://github.com/appbaseio/reactivebase/issues/210) Support native events for Input fields


app/sensors/CategorySearch.js
```diff
@@ -446,11 +446,14 @@ export default class CategorySearch extends Component {
 				value: this.state.currentValue.label
 			});
 		}
+		if (this.props.onBlur) this.props.onBlur(event);
 	}
 
 	handleKeyPress(event) {
 		if (event.key === "Enter") {
 			event.target.blur();
+		} else {
```
metagrover: Why do we have an `else` block here? Any good reasons?

dhruvdutt: I thought the other way around. But, yeah they should always be executed when present.

dhruvdutt:
Updated.

PR #89, opened 2017-09-12
Change prop componentStyle to style, fix scope

dhruvdutt:
- Change prop `componentStyle` to `style`
- Change actuators `style` prop scope to outer container

### DIFF
--- a/app/actuators/ResultCard.js
+++ b/app/actuators/ResultCard.js
@@ -691,7 +691,8 @@ ResultCard.propTypes = {
 	react: React.PropTypes.object,
 	scrollOnTarget: React.PropTypes.object,
 	pages: React.PropTypes.number,
-	pageURLParams: React.PropTypes.bool
+	pageURLParams: React.PropTypes.bool,
+	className: React.PropTypes.string
 };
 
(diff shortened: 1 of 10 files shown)


app/actuators/ResultCard.js
```diff
@@ -627,8 +626,8 @@ export default class ResultCard extends Component {
 		}
 
 		return (
-			<div className="rbc rbc-resultcard" ref={(node) => { this.listContainer = node; }}>
-				<div ref={(div) => { this.listParentElement = div; }} className={`rbc-resultcard-container card thumbnail ${cx}`} style={this.props.componentStyle} style={this.getComponentStyle()}>
```
dhruvdutt: There was a little glitch about `this.props.componentStyle` being used here and also inside `getComponentStyle()` method.

metagrover: You are merging `style` and `this.props.style` here, that shouldn't happen. You should keep both separate and add `this.props.style` at the root level

dhruvdutt:
The duplication of `style` prop was present in the old code itself. It's fixed now.

PR #90, opened 2017-09-12
feat: className prop support

dhruvdutt:


app/actuators/ResultCard.js
```diff
@@ -627,7 +627,7 @@ export default class ResultCard extends Component {
 		}
 
 		return (
-			<div className="rbc rbc-resultcard" ref={(node) => { this.listContainer = node; }}>
+			<div className=`rbc rbc-resultcard ${this.props.className ? this.props.className : ""}` ref={(node) => { this.listContainer = node; }}>
```
metagrover: You need to wrap className values with`{ ... }`.

dhruvdutt: Updated.
Thank you for fixing it in reactivebase. :sweat_smile:

PR #91, opened 2017-09-13
feat: onData prop className support

dhruvdutt:
Support for `onData` prop object - custom `className` property that gets injected with: 
- ResultList: `class="rbc-resultcard-item custom-class"`
- ResultCard: `class="rbc-resultlist-item custom-class"`

Example usage:
```jsx
  onData={
    function(res) {
      return {
        image: res.image,
        title: res.name,
        className: "custom-class",
        desc: (
            <div>
                <div className="price">${res.price}</div>
                <p>{res.room_type} · {res.accommodates} guests</p>
            </div>
        ),
        url: res.listing_url
      }
    }
  }
```

### DIFF
--- a/app/actuators/ResultCard.js
+++ b/app/actuators/ResultCard.js
@@ -336,7 +336,7 @@ export default class ResultCard extends Component {
 		markup = data.map((item) => {
 			const result = this.props.onData(item._source);
 			let cx = result.image ? "rbc-image-active" : "rbc-image-inactive";
-			cx = result.className ? cx + " " + result.className : cx;
+			cx = `${cx} ${result.className ? result.className : ""}`;
 			const details = (
 				<div>
 					{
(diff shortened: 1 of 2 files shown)


app/actuators/ResultCard.js
```diff
@@ -335,7 +335,8 @@ export default class ResultCard extends Component {
 
 		markup = data.map((item) => {
 			const result = this.props.onData(item._source);
-			const cx = result.image ? "rbc-image-active" : "rbc-image-inactive";
+			let cx = result.image ? "rbc-image-active" : "rbc-image-inactive";
+			cx = result.className ? cx + " " + result.className : cx;
```
metagrover: Cleaner way to write this:
```
cx = `${cx} ${result.className ? result.className : ""}`;
```

dhruvdutt: Cleaned up and updated. :+1:

PR #99, opened 2017-09-26
Add autoFocus support for input fields

dhruvdutt:


app/sensors/CategorySearch.js
```diff
@@ -553,6 +553,7 @@ export default class CategorySearch extends Component {
 								onKeyPress={this.props.onKeyPress}
 								onKeyDown={this.props.onKeyDown}
 								onKeyUp={this.props.onKeyUp}
+								autoFocus={this.props.autoFocus ? this.props.autoFocus : false}
```
metagrover: You can define `autoFocus` defaultProp as `false` and simply do:
```
autoFocus: {this.props.autoFocus}
```

PR #190, opened 2018-01-08
Fix examples anchor tag

john-d-pelingo:


metagrover:
Thanks @john-d-pelingo 💯

Issue #216, opened 2018-01-18
native: Enhancements / Bugs


dhruvdutt:
**Issue Type:**
Enhancements / Bugs

**Description:**
- ~[ ] Rename components~
	- ~SingleDropdownList / MultiDropdownList => SingleList / MultiList~
	- ~SingleDropdownRange / MultiDropdownRange => SingleRange / MultiRange~
- [x] Ignore the bucket key with an empty or whitespace value.
- [ ] Change the dir structure/move sensor components to relevant folders.
    - [ ] Add MultiDropdown<Range,Slider>
- [x] RangeSlider not working.

**Reactivesearch version:** latest

metagrover:
Added RangeSlider component - f2ae670b90e2e18f90cb941eec525a25f11b31b3

PR #224, opened 2018-01-22
native: Ignore the bucket key with an empty or whitespace value

dhruvdutt:
#216

### DIFF
--- a/packages/native/src/components/list/SingleDropdownList.js
+++ b/packages/native/src/components/list/SingleDropdownList.js
@@ -194,7 +194,7 @@ class SingleDropdownList extends Component {
 				{
 					[
 						...selectAll,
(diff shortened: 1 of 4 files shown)


packages/native/src/components/list/SingleDropdownList.js
```diff
@@ -192,7 +192,10 @@ class SingleDropdownList extends Component {
 				style={this.props.style}
 			>
 				{
-					[...selectAll, ...this.state.options].map(item => (
+					[
+						...selectAll,
+						...this.state.options.filter(item => String(item.key).length),
```
divyanshu013: Using `String(item.key).trim().length` should give better results

metagrover: yes, use `.trim()`

PR #228, opened 2018-01-23
Fixed typo

mauricedb:


metagrover:
Hi @mauricedb, thanks for taking out time to send this PR. 

As per urban dictionary, [Moar](https://www.urbandictionary.com/define.php?term=moar) is a fancy combination of "more" and "roar".  🐱  So, it wasn't a typo 😅

mauricedb:
LOL never heard that before. Learned a new thing today 😀

siddharthlatest:
@metagrover perhaps we should emphasize the word moar different (font weight, type, emphasis), something subtle so as to not distract but also show this was purposefully kept.

PR #229, opened 2018-01-23
native: DataSearch example

dhruvdutt:


metagrover:
I've some feedback for you:

1. Navigation-drawer isn't working properly. Seeing this error:

screen shot 2018-01-24 at 3 44 53 pm

2. Top navigation bar arrangement

screen shot 2018-01-24 at 3 45 12 pm

- You can use better font colors (white would enable great contrast). 
- "DataSearch" title should be in center.
- Status bar color should also be white.

dhruvdutt:
Made the fixes, I'm not sure about the status bar for iOS. I don't have any iOS devices ATM.

dhruvdutt:
@metagrover Can you check now? Everything should work except the back button on ios.
We've [implemented](https://github.com/dhruvdutt/reactivesearch/blob/b21345f52ae2f240ba6d13c5757dbf93e5842679/packages/native/examples/DataSearch/src/DataSearchView.js#L91) it in [correct way](https://docs.expo.io/versions/latest/guides/linking.html#in-the-expo-client), seems to be some issue with Expo ATM. The app is also live at [https://exp.host/@dhruvdutt/datasearch](url).

PR #249, opened 2018-02-06
native: Date<x> fix reset button bug

dhruvdutt:
Before:
image

After:
image


packages/native/src/components/date/DatePicker.js
```diff
@@ -261,7 +261,10 @@ class DatePicker extends Component {
 											{...getInnerKey(this.props.innerProps, 'button')}
 										>
 											<Text
-												style={getInnerKey(this.props.innerStyle, 'label')}
+												style={{
+													color: this.props.theming.primaryTextColor,
```
metagrover: This won't work in ios. You should conditionally apply `color: this.props.theming.primaryTextColor` for Android

PR #275, opened 2018-02-17
web: Update & fix prop types in all components

dhruvdutt:
- Separation of props binded from `mapDispatchtoProps`, `mapStateToProps` and component props (This will make managing props that are exposed by component easier) 
- Fixes missing propsTypes at multiple places


metagrover:
@dhruvdutt Thanks for contributing 💯

Issue #291, opened 2018-03-13
Feature: Support aggregating on missing value for the list components


lopper:
**Issue Type:**
enhancement

**Description:**
It would be nice to have a way to specify that the `SingleList`, `MultiList`, `SingleDropdownList`, and `MultiDropdownList` should return an entry for items that do not have a value.  Specifically, I want to be able to add the "missing" property to the terms aggregation query.  

We could add the following two properties to each of the components listed above.
`showMissing:<boolean>`
`missingLabel:<string>`

davidklebanoff:
This is somewhat related to discussions in #269.

davidklebanoff:
@lopper et al. - perhaps a means to disable the the missing items? So they are shown but not interactive, seeing as clicking one would be a no-op.

lopper:
@davidklebanoff I think you have misinterpreted what I would like to do. Let me give an example. 

If I have the following 5 items in elasticsearch. 
```
{_id: 1, tag : 'apple'}
{_id: 2, tag : 'orange'}
{_id: 3, tag : 'orange'}
{_id: 4, tag : 'grape'}
{_id: 5 }
```
Right now, the `SingleList` with the count option enable would show in the list
- apple (1)
- grape (1)
- orange (2)

I want to have a way to add an additional entry so that I can filter the search results with items that have no tag entry. I would like the list to show

- apple (1)
- grape (1)
- orange (2)
- N/A (1)

My thinking is that setting `showMissing` to true could indicate to change the aggregate query for filter options to return a count for the items that are missing the dataField. The `missingLabel` would determine the label that is used to represented the items with the missing dataField.

davidklebanoff:
Ah yes, that makes more sense. I thought you may have meant _missing_ as in returning a 0 for the current filters.

metagrover:
TODO:

- [ ] Add docs
- [ ] Support MultiDropdownList and MultiList

divyanshu013:
Hi @lopper, do you want to send PR for MultiDropdownList and MultiList too?

lopper:
@divyanshu013 Yes, I'll send a PR for MultiDropdownList and MultiList.

PR #293, opened 2018-03-14
web: support filtering by items with missing property in SingleDropdownList and SingleList.

lopper:
Implementation for #291. I have only implemented the suggested feature for SingleDropdownList and SingleList. I have not yet implemented it for MultiDropdownList and MultiList.

### DIFF
--- a/packages/web/src/components/list/SingleDropdownList.js
+++ b/packages/web/src/components/list/SingleDropdownList.js
@@ -107,8 +107,8 @@ class SingleDropdownList extends Component {
 	};
 
 	defaultQuery = (value, props) => {
-		if (this.props.selectAllLabel && this.props.selectAllLabel === value) {
-			if (this.props.showMissing === true) {
+		if (props.selectAllLabel && props.selectAllLabel === value) {
+			if (props.showMissing === true) {
 				return { match_all: {} };
 			}
 			return {
@@ -117,7 +117,7 @@ class SingleDropdownList extends Component {
 				},
 			};
 		} else if (value) {
-			if (this.props.showMissing === true && value === this.props.missingLabel) {
+			if (props.showMissing && value === props.missingLabel) {
 				return {
 					bool: {
 						must_not: {
@@ -184,7 +184,7 @@ class SingleDropdownList extends Component {
 					field: props.dataField,
 					size: props.size,
 					order: getAggsOrder(props.sortBy),
-					...(this.props.showMissing ? { missing: this.props.missingLabel } : {}),
+					...(props.showMissing ? { missing: props.missingLabel } : {}),
 				},
 			},
 		};
--- a/packages/web/src/components/list/SingleDropdownList.js
+++ b/packages/web/src/components/list/SingleDropdownList.js
@@ -108,12 +108,24 @@ class SingleDropdownList extends Component {
 
 	defaultQuery = (value, props) => {
 		if (this.props.selectAllLabel && this.props.selectAllLabel === value) {
+			if (this.props.showMissing === true) {
+				return {};
+			}
 			return {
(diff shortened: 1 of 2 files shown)


packages/web/src/components/list/SingleDropdownList.js
```diff
@@ -108,12 +108,24 @@ class SingleDropdownList extends Component {
 
 	defaultQuery = (value, props) => {
 		if (this.props.selectAllLabel && this.props.selectAllLabel === value) {
+			if (this.props.showMissing === true) {
+				return {};
```
metagrover: I think it would be better if we return a `{ match_all: {} }` query here, instead of `{}`, allowing us to clearly state the intent that we expect all the results from this component's query.

Also, notice how we pass `props` as a param to this method. So, you should rewrite the boolean condition here as:

```
if (props.showMissing)
```

packages/web/src/components/list/SingleDropdownList.js
```diff
@@ -108,12 +108,24 @@ class SingleDropdownList extends Component {
...
+			}
 			return {
 				exists: {
 					field: props.dataField,
 				},
 			};
 		} else if (value) {
+			if (this.props.showMissing === true && value === this.props.missingLabel) {
```
metagrover: Again, you can simplify the boolean condition here:

```
if (props.showMissing && value === props.missingLabel)
```

packages/web/src/components/list/SingleDropdownList.js
```diff
@@ -172,6 +184,7 @@ class SingleDropdownList extends Component {
 					field: props.dataField,
 					size: props.size,
 					order: getAggsOrder(props.sortBy),
+					...(this.props.showMissing ? { missing: this.props.missingLabel } : {}),
```
metagrover: Refrain from using `props` from the context wherever possible. Since we call these functions from the lifecycle methods like `componentWillReceiveProps` where the updated props are not yet available on `this.props`. Hence, you should use `props` object coming from the params.

packages/web/src/components/list/SingleList.js
```diff
@@ -110,12 +110,24 @@ class SingleList extends Component {
 
 	defaultQuery = (value, props) => {
 		if (this.props.selectAllLabel && this.props.selectAllLabel === value) {
+			if (this.props.showMissing === true) {
```
metagrover: same as before

packages/web/src/components/list/SingleList.js
```diff
@@ -110,12 +110,24 @@ class SingleList extends Component {
...
+				return {};
+			}
 			return {
 				exists: {
 					field: props.dataField,
 				},
 			};
 		} else if (value) {
+			if (this.props.showMissing === true && value === this.props.missingLabel) {
```
metagrover: same as before

packages/web/src/components/list/SingleList.js
```diff
@@ -179,6 +191,7 @@ class SingleList extends Component {
 					field: props.dataField,
 					size: props.size,
 					order: getAggsOrder(props.sortBy),
+					...(this.props.showMissing ? { missing: this.props.missingLabel } : {}),
```
metagrover: same as before

metagrover:
Hello @lopper 👋.  Thanks for contributing. I've requested some changes and it'd be awesome if you can update the code. Feel free to ask me any questions regarding the same.

Cheers!

lopper:
@metagrover I've made the changes you requested.

metagrover:
@lopper Thank you for the contribution 🎉 

This enhancement will be out in the next release (along with the changes for MultiDropdownList and MultiList).

PR #296, opened 2018-03-15
web: support filtering items with a missing property in MultiDropdownList and MultiList

### LINKED ISSUE #291: Feature: Support aggregating on missing value for the list components

lopper:
Implementation for #291.

Resolves #291


divyanshu013:
Thanks for contributing @lopper. We'll roll this out in the upcoming release. Cheers! 🍻

Issue #287, opened 2018-03-09
Feature Request: support making nextPage, prevPage, and setPage functions public on reactiveList.


lopper:
**Issue Type:**
enhancement

**Description:**
It would be nice to be able to programmatically move to a particular page of the search result. Right now, the  paging options can only be evoked by user input. Is there a way to expose the nextPage, prevPage, and setPage functions of the reactiveList?

lopper:
@divyanshu013 are there any thoughts on this enhancement?

metagrover:
For this use-case, we will support the following API:

- `currentPage` prop - which can be used to set and update on the current page of the result component.
- `onPageChange` prop - will be enhanced to expose the current page number on page change.

lopper:
@metagrover nice. I will try to make a pull request for this change. One question, what do you think should hold precedent, the currentPage prop or the page query parameter when the component is first loaded?

siddharthlatest:
@lopper iirc, we give precedence to the URLParam value if set on initial load (a user may be opening a saved URL from before), and thereafter the value set via the prop (as it may be programmatically changing)

lopper:
@siddharthlatest how do you keep track of the state of whether the component has finished initial load in the mapStateToProps? 

The currentPage property current reads it's value from the store. Here is the current code.
```
currentPage: (
		state.selectedValues[`${props.componentId}-page`]
		&& state.selectedValues[`${props.componentId}-page`].value - 1
	) || 0,
```
I'm assuming we need to do something like.

```
currentPage: (
		!initLoad && state.selectedValues[`${props.componentId}-page`]
		&& state.selectedValues[`${props.componentId}-page`].value - 1
	) || this.props.currentPage || 0,
```
I'm thinking the only way to track an initLoad value would be to create another reduced in reactiveCore. However, I don't think adding another reducer to accomplish this would be very good. It would add a lot more code.

However, if we make the requirement so that the URLParam does not hold precedence, then the change will be much simpler. 

I can make the following change.
```
currentPage: (
	props.currentPage || state.selectedValues[`${props.componentId}-page`]
	&& state.selectedValues[`${props.componentId}-page`].value - 1
) || 0,
```
As long as the currentPage is not set as a prop on the component on initial load, the URLParam will be used to set the currentPage. Then, after initial load, the user can start setting the currentPage prop on the results component, which will allow changing the current page. 

Any thoughts on this?

metagrover:
[...]

PR #297, opened 2018-03-15
Support setting current page of result components from outside the component.

lopper:
This PR is for feature enhancement #287.

There are two main changes with this pull request. The first change allows the currentPage prop to define which page the result components will initially show. 

```
render() {
	<ReactiveList
		dataField="original_title.raw"
		size={2}
                 currentPage={5}
		pagination
		onData={this.booksReactiveList}
		react={{
			and: ['BookSensor'],
		}}
	/>
}
```
The second change provides a way to set the current page from outside of the result components after the result component has already been rendered. This second change will provide developers an avenue to implement custom pagination. 
Here is how it can be used.
```
onNextPage() => {
	const stats = this.reactiveList.getResultStats();
	this.reactiveList.setPage(stats.currentPage + 1};
}


render() {
	<ReactiveList
		componentId="SearchResult"
		dataField="original_title.raw"
		className="result-list-container"
		from={10}
		size={2}
		providerRef={(provider) => { this.reactiveList = provider; }}
		stream={false}
		pagination
		pages={5}
		paginationAt="bottom"
		onData={this.booksReactiveList}
		react={{
			and: ['BookSensor'],
		}}
	/>
}
```

### DIFF
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -471,7 +471,7 @@ const mapStateToProps = (state, props) => ({
 	currentPage: (
 		state.selectedValues[`${props.componentId}-page`]
 		&& state.selectedValues[`${props.componentId}-page`].value - 1
-	) || 0,
+	) || props.currentPage || 0,
 	hits: state.hits[props.componentId] && state.hits[props.componentId].hits,
 	isLoading: state.isLoading[props.componentId] || false,
 	streamHits: state.streamHits[props.componentId] || [],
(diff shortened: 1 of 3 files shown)


metagrover:
Hi @lopper, I reviewed this PR and I strongly feel that we can achieve the desired enhancements here simply by exposing the `currentPage` prop and enhancing the `onPageChange` prop. That way, we can avoid the `providerRef` exposure.

For #287 use-case, we will set/update the `currentPage` props in the result component. Here, we will be watching for the changes of `currentPage` prop and handle it accordingly in the component lifecycle method, and we can keep track of the active page via `onPageChange` callback with necessary enhancements.

Let me know if you want to take this up and implement it. I will be around to guide you through it if you need me to.

Closing this PR for now! Thanks for all the valuable contributions 👍

PR #300, opened 2018-03-17
Add cleartext  button to DataSearch

### LINKED ISSUE #255: Feature Request: Optional prop for a clear button on text/search components.
**Issue Type:**
Enhancement

**Description:**

image

Add a prop on the  TextField, DataSearch, and CategorySearch components which adds a clear button that clears all the text in the text input when clicked (and re-runs a search when appropriate). Similar to the [HTML 5 clear button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search) that would show for some browsers if type="search".

**Example Documentation:**

- **showClear** Boolean [optional] whether to display a clear or custom icon in the input box. Defaults to false.
- **clearIcon** HTML [optional] displays a custom clear icon instead of the default **X**

davkap92:
For #255


packages/web/src/components/search/DataSearch.js
```diff
@@ -378,6 +383,12 @@ class DataSearch extends Component {
 		return null;
 	}
 
+	renderCancelIcon = () => {
+		if (this.props.showClear) {
+			return <CancelSvg />;
+		}
```
divyanshu013: Use a prop `clearIcon` here if user wants to render a custom icon for clearing the field

packages/web/src/components/search/DataSearch.js
```diff
@@ -428,7 +439,13 @@ class DataSearch extends Component {
 										})}
 										themePreset={themePreset}
 									/>
+									{
```
divyanshu013: Instead of ternary syntax here `condition ? component : null` we can use `&&` to avoid the `else` part of returning `null`. For example:
```js
{
    this.state.currentValue && this.props.showClear && <Component />
}
```

divyanshu013:
Hi @davkap92, this looks good overall. Requested a couple of changes.

Currently it doesn't work when `autosuggest` prop is set to `false`. You can replicate the same behavior to resolve this issue.

Also try out an example with the `iconPosition` prop set to `right`. This is causing the search icon to overlap with the cancel icon.

davkap92:
@divyanshu013 Regarding the iconPosition right, more a design issue really.. but
we could do a check to only allow clearIcon if iconPosition is `left` ?

divyanshu013:
Merged changes via https://github.com/appbaseio/reactivesearch/commit/fb15d5494b30ed5cf2352b7f94a3668f6266641f https://github.com/appbaseio/reactivesearch/commit/4edb5c95444cbc8c28fb3339690ca276e5bebd3c

Thanks @davkap92 🔥. This will be out shortly 🙂

PR #302, opened 2018-03-20
Support setting the current page of the result component after it has been initialized. 

lopper:
Reference #287 

The following PR adds support to set the current page of the result component. Prior to the result component's initialization the current page can be set from the URLParam or the `currentPage` prop. The URLParam holds precedence over the currentPage prop. Once the initialization of the component is complete, the current page can be changed with the `currentPage` prop. 

In addition, this pull request modifies the `onPageChange` to export the `currentPage` and `totalPages` state variables.

### DIFF
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -168,7 +168,8 @@ class ReactiveList extends Component {
 				this.setState({
 					isLoading: false,
 				});
-			} else if (this.props.currentPage !== nextProps.currentPage
+			}
+			if (this.props.currentPage !== nextProps.currentPage
 					&& nextProps.currentPage > 0
 					&& nextProps.currentPage <= this.state.totalPages) {
 				this.setPage(nextProps.currentPage - 1);
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -176,7 +176,9 @@ class ReactiveList extends Component {
 				this.setState({
 					isLoading: false,
 				});
-			} else if (this.props.currentPage !== nextProps.currentPage) {
+			} else if (this.props.currentPage !== nextProps.currentPage
+					&& nextProps.currentPage > 0
+					&& nextProps.currentPage <= this.state.totalPages) {
 				this.setPage(nextProps.currentPage - 1);
 			}
 		}
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -166,7 +166,10 @@ class ReactiveList extends Component {
 		if (this.props.pagination) {
 			if (this.state.isLoading) {
 				if (nextProps.onPageChange) {
-					nextProps.onPageChange();
+					nextProps.onPageChange({
+						currentPage: this.state.currentPage + 1,
+						totalPages: this.state.totalPages,
+					});
 				} else {
 					window.scrollTo(0, 0);
 				}
@@ -199,7 +202,10 @@ class ReactiveList extends Component {
 			&& nextProps.hits.length < this.props.hits.length
 		) {
 			if (nextProps.onPageChange) {
-				nextProps.onPageChange();
+				nextProps.onPageChange({
+					currentPage: this.state.currentPage + 1,
+					totalPages: this.state.totalPages,
+				});
 			} else {
 				window.scrollTo(0, 0);
 			}
@@ -480,6 +486,7 @@ ReactiveList.defaultProps = {
 	size: 10,
 	style: {},
 	URLParams: false,
+	currentPage: 0,
 };
 
 const mapStateToProps = (state, props) => ({
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -30,24 +30,19 @@ class ReactiveList extends Component {
 	constructor(props) {
 		super(props);
 
+		let currentPage = 0;
+		if (this.props.defaultPage) {
+			currentPage = this.props.defaultPage - 1;
+		} else if (this.props.currentPage) {
+			currentPage = Math.max(this.props.currentPage - 1, 0);
+		}
+
 		this.state = {
 			from: props.currentPage * props.size,
 			isLoading: false,
-			currentPage: props.currentPage,
+			currentPage,
 		};
 		this.internalComponent = `${props.componentId}__internal`;
-	}
-
-	componentWillMount() {
-		if (this.props.defaultPage !== undefined) {
-			this.setState({
-				currentPage: this.props.defaultPage,
-			});
-		} else if (this.props.currentPage) {
-			this.setState({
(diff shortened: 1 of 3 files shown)


packages/web/src/components/result/ResultCard.js
```diff
@@ -41,6 +41,18 @@ class ResultCard extends Component {
 		this.internalComponent = `${props.componentId}__internal`;
 	}
 
+	componentWillMount() {
```
metagrover: It'd be best if we do this evaluation in the constructor and kill the unnecessary noise.

packages/web/src/components/result/ResultCard.js
```diff
@@ -154,15 +166,19 @@ class ResultCard extends Component {
...
 		// called when page is changed
-		if (this.props.pagination && this.state.isLoading) {
-			if (nextProps.onPageChange) {
-				nextProps.onPageChange();
-			} else {
-				window.scrollTo(0, 0);
+		if (this.props.pagination) {
+			if (this.state.isLoading) {
+				if (nextProps.onPageChange) {
+					nextProps.onPageChange();
```
metagrover: `onPageChange()` callback function should have a param - current page number.

packages/web/src/components/result/ResultCard.js
```diff
@@ -154,15 +166,19 @@ class ResultCard extends Component {
...
+				} else {
+					window.scrollTo(0, 0);
+				}
+				this.setState({
+					isLoading: false,
+				});
+			} else if (this.props.currentPage !== nextProps.currentPage) {
```
metagrover: why do we have this condition in an `else if`? 🤔

packages/web/src/components/result/ResultCard.js
```diff
@@ -497,13 +514,14 @@ ResultCard.defaultProps = {
...
+	currentPage: 0,
 };
 
 const mapStateToProps = (state, props) => ({
-	currentPage: (
+	defaultPage: (
 		state.selectedValues[`${props.componentId}-page`]
 		&& state.selectedValues[`${props.componentId}-page`].value - 1
-	) || 0,
+	),
```
metagrover: must have `0` as a default value. `(...) || 0` should fix it

packages/web/src/components/result/ReactiveList.js
```diff
@@ -152,15 +164,24 @@ class ReactiveList extends Component {
...
 		// called when page is changed
-		if (this.props.pagination && this.state.isLoading) {
-			if (nextProps.onPageChange) {
-				nextProps.onPageChange();
-			} else {
-				window.scrollTo(0, 0);
+		if (this.props.pagination) {
+			if (this.state.isLoading) {
+				if (nextProps.onPageChange) {
+					nextProps.onPageChange({
```
metagrover: We can clean up the callback param here. Pass `currentPage` and `totalPages` as two params rather than an object.

packages/web/src/components/result/ReactiveList.js
```diff
@@ -152,15 +164,24 @@ class ReactiveList extends Component {
...
+						currentPage: this.state.currentPage + 1,
+						totalPages: this.state.totalPages,
+					});
+				} else {
+					window.scrollTo(0, 0);
+				}
+				this.setState({
+					isLoading: false,
+				});
+			} else if (this.props.currentPage !== nextProps.currentPage
```
metagrover: Why do we have this in an `else if` rather than having it as an independent `if` block? 🤔

packages/web/src/components/result/ReactiveList.js
```diff
@@ -195,10 +219,19 @@ class ReactiveList extends Component {
...
+			const currentPage = this.props.total ? 0 : this.state.currentPage;
 			this.setState({
-				totalPages: Math.ceil(nextProps.total / nextProps.size),
-				currentPage: this.props.total ? 0 : this.state.currentPage,
+				totalPages,
+				currentPage,
 			});
+
+			if (this.props.onPageChange) {
+				this.props.onPageChange({
```
metagrover: params cleanup!

packages/web/src/components/result/ResultCard.js
lopper: @metagrover I made the `defaultPage` undefinable so that I could use it to detect whether `defaultPage` has been set from the url query value and, if not set, then ignore it and prefer the use of the `currentPage` prop on component initialization. The problem is that when the `defaultPage` can be set with a default value of 0, I cannot determine whether the reason for the value is because the url query value is 1 or whether the `defaultPage` defaulted to 0.  I need a way on initialization to determine whether the `defaultPage` should be ignored or used.

To allow the defaultPage prop to default to 0, I believe I can change the code to make the `defaultPage` valid values start at 1 and interpret 0 as ignore. To accomplish this, I can change the `defaultPage` portion of code in `mapStateToProps` to remove the subtract 1 from the selectedValue and set the default value to 0.  Here is what the code can be:

```
defaultPage: (
    state.selectedValues[`${props.componentId}-page`]
	&& state.selectedValues[`${props.componentId}-page`].value
) || 0,
```

metagrover: I agree, maybe we can do `-1` instead of `undefined`.

packages/web/src/components/result/ReactiveList.js
```diff
@@ -496,8 +491,8 @@ ReactiveList.defaultProps = {
 const mapStateToProps = (state, props) => ({
 	defaultPage: (
 		state.selectedValues[`${props.componentId}-page`]
-		&& state.selectedValues[`${props.componentId}-page`].value - 1
-	),
+		&& state.selectedValues[`${props.componentId}-page`].value
+	) || 0,
```
metagrover: You were right, we need an 'invalid' value here to determine the cases where defaultPage is not set. We can use `-1` here, to indicate the same. What are your thoughts?

packages/web/src/components/result/ResultCard.js
```diff
@@ -32,27 +32,22 @@ class ResultCard extends Component {
 	constructor(props) {
 		super(props);
 
+		let currentPage = 0;
+		if (this.props.defaultPage) {
```
metagrover: we can check here if the defaultPage >= 0 to be valid. (after setting it as -1 as the invalid value)

packages/web/src/components/result/ReactiveList.js
metagrover: cc @lopper

packages/web/src/components/result/ResultCard.js
lopper: @metagrover Sounds good. I'll make the change.

metagrover:
PR Merged here: https://github.com/appbaseio/reactivesearch/commit/16f84fda590dd3bfafcd9de898c24d6f6df38fd7 (Not sure what's wrong with Github)

Thanks for contributing 🎉

Please check the updates I made on this PR to get it to work properly - https://github.com/appbaseio/reactivesearch/commit/35f251377c74a325b62265bbd111a20b2326f3a2

Issue #298, opened 2018-03-16
Pagination doesn't correctly update when the size property in the Results is changed


lopper:
**Issue Type:**
bug

**Platform:**
Web

**Description:**
When I change the size property of the ReactiveList, the pagination component does not update to reflect the correct last possible page. 

 For example, if I start with a ReactiveList with a pagination size of 10 and there are 300 results, the last page is page 30. However, if I then change the size to 50, the pagination component does not update to reflect that the last page should now be 6. The pagination still shows 30 as the last page.

siddharthlatest:
Thanks for filing this issue @lopper! We will try to incorporate this for the next release. If you have already identified the root cause - feel free to send a PR.

PR #304, opened 2018-03-21
Update currentPage and totalPages in result components when page size changes.

lopper:
PR for #298.
Recalculate the currentPage and totalPages when the the size prop of the result components is changed.


metagrover:
Hi @lopper, this is not a valid change. Since it will only update the the pagination component (number of pages) and not the number of results being displayed currently. When the `size` prop is changed, new query should be generated which will handle the results and the number of pages in the pagination component accordingly with the new set of hits.

Issue #284, opened 2018-03-07
In DynamicRange Silder Tooltip is not present in V2.0.0 


ashwini987:
**Issue Type:**
 enhancement' issue. 

**Platform:**
Web

**Description:**
In DynamicRange Silder Tooltip is not there, So how can user know that which range is selected ??? 
In V1 version tooltip is present. but not in v2.0.0

**Screenshots:**
![dynamic-tooltip](https://user-images.githubusercontent.com/31088567/37074003-5b2d51fc-21f0-11e8-9254-c1e50c3a97a6.png)

**Reactivesearch version:** 
https://github.com/appbaseio/reactivesearch/releases/tag/v2.0.0  ------> I m using this version right now 

**Browser:** [all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

divyanshu013:
We use `rheostat` from `airbnb` for the slider component. Currently there is no functionality to show tooltips in rheostat. One way to add this functionality would be to send a PR to `rheostat`. [Here](https://github.com/airbnb/rheostat/issues/134) is the related issue.

Another alternative if someone wishes to show the current slider positions would be to expose a prop which returns the current values of the slider (before running the query) and use them in the UI just like airbnb does. Ref-

![](https://screenshots.firefox.com/RGhkI0Ivainy4v7C/www.airbnb.co.in)

metagrover:
I think the solution mentioned in [this comment](https://github.com/airbnb/rheostat/issues/134#issuecomment-358129687) seems relevant. We will try it out internally, and if it works, we will roll out an updated for this.

Will keep the progress posted here!

ashwini987:
for this issue, i am using  **"title" : y[n]**  in Build File(index.js), so atleast user know price not on current slider positions but at steady position.

PR #305, opened 2018-03-21
Expose prop callback for slider values

davkap92:
For issue #284 
Doesn't solve the tooltip part but exposes the value so can at least do what AirBnb does
i.e the rangeLabels update with the slider values

### DIFF
--- a/packages/web/src/components/range/DynamicRangeSlider.js
+++ b/packages/web/src/components/range/DynamicRangeSlider.js
@@ -223,6 +223,9 @@ class DynamicRangeSlider extends Component {
 			this.setState({
 				currentValue: normalizedValue,
 			}, () => {
+				if (this.props.currentValues) {
+					this.getSliderValues([normalizedValue[0], normalizedValue[1]], props);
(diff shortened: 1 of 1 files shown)


packages/web/src/components/range/DynamicRangeSlider.js
```diff
@@ -223,6 +223,9 @@ class DynamicRangeSlider extends Component {
 			this.setState({
 				currentValue: normalizedValue,
 			}, () => {
+				if (this.props.currentValues) {
+					this.getSliderValues([normalizedValue[0], normalizedValue[1]], props);
```
metagrover: Hi @davkap92, I don't think this is an appropriate way to expose slider values.

davkap92:
Valid points 👍  Sure will have a look soon, was trying to get the realtime values during slide in handleChange but perhaps too late by that stage. Thanks

davkap92:
ok amended, not sure why need a new pr?, not that different unless I'm missing something :)

Issue #313, opened 2018-03-27
setAggs Reactive Component


justinsoong:
Upon trying to use Reactive Component SetQuery, realising that it only sets the query attribute of the whole ES query.

I want to able to change the aggs attribute as well. Maybe something called setAggs?

I can look into it if no one else has?

metagrover:
Hi @justinsoong! Thank you for filing this issue.

I just realised that we don't support it via `setQuery` atm. But you should be able to set the aggregations query via `aggs` key in the `defaultQuery`. Check out this [example](https://github.com/appbaseio/reactivesearch/blob/dev/packages/web/examples/ReactiveComponent/src/index.js#L26-L38), and since we watch for the changes on the `defaultQuery` prop, any change in the `defaultQuery()` will trigger a new request.

It'd be awesome if you can take this up and expose a `setAggs` method for the same. [This](https://github.com/appbaseio/reactivesearch/blob/dev/packages/web/src/components/basic/ReactiveComponent.js#L51-L64) should help you get started.

Feel free to reach out to me if you any questions 😄

justinsoong:
@metagrover 
Cool having a look, but noticed that both componentWillMount where defaultQuery first called, using updateQuery via query key, don't see how this is any different from setQuery looks the same...apart from the componentId key, anyway i just forked, will dig deeper

any change in the defaultQuery() will trigger a request, when you mean change, here zoom changes, via setState in the Component, im not observing a second request.
```
defaultQuery={() => {
                return {
                  aggs: {
                    two: {
                      geohash_grid: {
                        field: "location",
                        precision: zoom
                      },
                      aggs: {
                        three: {
                          geo_centroid: {
                            field: "location"
                          }
                        }
                      }
                    }
                  }
                };
              }}
```

metagrover:
I just checked, you're right! The dynamic `defaultQuery` behavior is not supported in ReactiveComponent. If you wish to work on this - you can get started by watching for the `defaultQuery()` changes in the lifecycle method - ex: [check the componentWillReceiveProps method here](https://github.com/appbaseio/reactivesearch/blob/dev/packages/web/src/components/result/ReactiveList.js#L133-L136), where we handle the changes on `defaultQuery`.

And, maybe instead of exposing a `setAggs()` method, we can update the `setQuery()` method with the standard keys for `query` and `aggs` in place and handling them accordingly.

justinsoong:
[...]

PR #316, opened 2018-03-28
defaultQuery watch on change for ReactiveComponent

justinsoong:
#313

### DIFF
--- a/packages/web/src/components/basic/ReactiveComponent.js
+++ b/packages/web/src/components/basic/ReactiveComponent.js
@@ -11,6 +11,7 @@ import {
 	pushToAndClause,
 	parseHits,
 	isEqual,
+	getQueryOptions,
 } from '@appbaseio/reactivecore/lib/utils/helper';
 import types from '@appbaseio/reactivecore/lib/utils/types';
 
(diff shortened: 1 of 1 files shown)


packages/web/src/components/basic/ReactiveComponent.js
```diff
@@ -73,6 +74,27 @@ class ReactiveComponent extends Component {
...
+		) {
+			const options = getQueryOptions(nextProps);
+			options.from = this.state.from;
+			this.defaultQuery = nextProps.defaultQuery();
+
+			const { sort, ...query } = this.defaultQuery;
+
+			if (sort) {
+				options.sort = this.defaultQuery.sort;
+				nextProps.setQueryOptions(nextProps.componentId, options, !query);
```
metagrover: I don't think this will work as expected. `defaultQuery()` in case of ReactiveComponent should be executed on the internal component (for both `updateQuery()` and `setQueryOptions()`).

metagrover: Also, you should add support to execute query on any change in `queryOptions`, not just `sort`. It gets handled in a different manner in other components, but that's not the case here since this component is too vanilla as per its specs.

metagrover:
Closing due to inactivity. Check the implementation [here](https://github.com/appbaseio/reactivesearch/commit/2ef9941b703cdbdf9bb265589d3a3d2c1fedd866).

Issue #223, opened 2018-01-21
Typescript support


yytsui:
**Issue Type:enhancement**

**Description:**

Any plan to support  [type definition file](https://github.com/DefinitelyTyped/DefinitelyTyped) for Typescript?
**Browser:** [all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

**Anything else:**

Thanks!

romainquellec:
> 2.3.0, someday

It's time ! (Will see how to do it)

metagrover:
That'd be awesome. We have been swamped with the stuff on `reactivemaps` among other things 😅 

Feel free to reach out to me here or on [gitter](https://gitter.im/appbaseio/reactivesearch) if you need any help or suggestions. Looking forward to see the typescript support 😄

romainquellec:
@yytsui, do you know how to make it ?

romainquellec:
In short, we need something like `import * as React from 'react';

declare namespace DataController {
    export interface DataControllerProps {
      addComponent: void,
    	removeComponent: void,
    	updateQuery: void,
    	selectedValue?: any,
    	componentId: string,
    	beforeValueChange?: void,
    	children?: any,
    	className?: string,
    	customQuery?: void,
    	defaultSelected?: any,
    	filterLabel?: string,
    	onQueryChange?: void,
    	onValueChange?: void,
    	showFilter?: boolean,
    	style?: any,
    	URLParams: boolean,
    }
    //export interface DataControllerState {}
}
//declare class DataController extends React.Component<DataController.DataControllerProps, DataController.DataControllerState> { }
declare class DataController extends React.Component<DataController.DataControllerProps> { }
export default DataController;`

For each components.

metagrover:
Yeah, that's pretty much it. I agree it is quite a bit of work 😅 

We have been looking into [material-ui](https://github.com/mui-org/material-ui/tree/v1-beta/src/Tabs)'s type definitions and we will probably use a similar pattern. 

Here's how I'm thinking of achieving this: Start with defining a standard component type definitions (should be simple, since most of the components share same method naming) and then extending that definition to create typings for every component.

romainquellec:
Well, I don't have so much knowledge about your components. How can we work on this ?

**I'm proposing this :** 
[...]

PR #325, opened 2018-04-11
web: Add typescript interfaces, types for components

dhruvdutt:
image 1

#223

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,4 +6,6 @@ playground
 packages/web/webpack.config.js
 packages/web/webpack.config.umd.js
 packages/maps/webpack.config.umd.js
+packages/web/src/types.js
+packages/web/src/types.js.map
 .expo/
--- a/packages/web/src/components/basic/NumberBox.d.ts
+++ b/packages/web/src/components/basic/NumberBox.d.ts
@@ -1,0 +1,18 @@
+import * as React from "react";
+import { CommonProps } from "../../";
+import * as types from "../../types";
+
+export interface NumberBoxProps extends CommonProps {
+	data?: types.dataNumberBox;
+	dataField?: string;
+	defaultSelected?: number;
+	innerClass?: types.style;
+	labelPosition?: types.labelPosition;
+	queryFormat?: types.queryFormatNumberBox;
+	react?: types.react;
+	title?: types.title;
+}
+
+declare const NumberBox: React.ComponentType<NumberBoxProps>;
+
+export default NumberBox;
--- a/packages/web/src/types.ts
+++ b/packages/web/src/types.ts
@@ -1,0 +1,190 @@
+export type components = string[];
+
+export type children = any;
+
+export type data = any[];
+
+export type dataFieldArray =
+	| string
+	| string[];
+
+export interface dataNumberBox {
+	label?: string;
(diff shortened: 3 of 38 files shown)


.eslintignore
```diff
@@ -6,4 +6,6 @@ playground
 packages/web/webpack.config.js
 packages/web/webpack.config.umd.js
 packages/maps/webpack.config.umd.js
+packages/web/src/types.js
```
divyanshu013: Why are we ignoring linting here?

packages/web/src/types.ts
```diff
@@ -0,0 +1,190 @@
...
+	end: number;
+}
+
+export type date =
+	| string
+	| string[];
+
+export type dateObject = object;
+
+export interface dateRange {
```
divyanshu013: What is this being used for?

packages/web/src/types.ts
```diff
@@ -0,0 +1,190 @@
...
+	| object
+	| object[];
+
+export type paginationAt =
+	| "top"
+	| "bottom"
+	| "both";
+
+export interface range {
+	start?: number;
```
divyanshu013: `start` and `end` are required keys

packages/web/src/types.ts
```diff
@@ -0,0 +1,190 @@
...
+	end?: number;
+}
+
+export interface rangeLabels {
```
divyanshu013: `start` and `end` should be required keys

packages/web/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+    "compilerOptions": {
+      "module": "commonjs",
+      "target": "es6",
+      "lib": ["es6", "dom"],
+      "sourceMap": true,
+      "jsx": "react",
+      "moduleResolution": "node",
+    //   "noImplicitAny": true,
```
divyanshu013: If these are not required, we can remove them

packages/web/src/components/basic/NumberBox.d.ts
```diff
@@ -0,0 +1,18 @@
+import * as React from "react";
+import { CommonProps } from "../../";
+import * as types from "../../types";
+
+export interface NumberBoxProps extends CommonProps {
+	data?: types.dataNumberBox;
+	dataField?: string;
+	defaultSelected?: number;
+	innerClass?: types.style;
```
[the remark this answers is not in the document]
metagrover: I think it's fine as it is. Otherwise, we will have to define it for all the components and it vary from component to component.

packages/web/src/types.ts
dhruvdutt: Good catch. This has been mirrored from core.
I think it's unused we can clean it up from both places.

packages/web/src/types.ts
dhruvdutt: Mirrored from core/types.js
Should I make it required at both places?

divyanshu013: Yeah

packages/web/tsconfig.json
dhruvdutt: This is a good config to follow but we might not comply all rules so better to clean it up. :+1:

PR #327, opened 2018-04-11
web: Fix onValueChange behavior in all components

dhruvdutt:
Also, check web: RangeInput

https://github.com/appbaseio/reactivesearch/blob/cfc2db1404bc893e0d90d47140d46f70376e3c9a/packages/web/src/components/range/RangeInput.js#L69-L80


metagrover:
@dhruvdutt Looks good!

The logic for handling `onValueChange` in RangeInput is correct (as it uses `onValueChange` of internal `RangeSlider`).

PR #328, opened 2018-04-11
native: Fix onValueChange behavior in all components

dhruvdutt:


packages/native/src/components/date/DateRange.js
```diff
@@ -193,9 +193,10 @@ class DateRange extends Component {
 			this.setState({
 				currentDate,
 			}, () => {
+				this.updateQuery(value, props);
+				if (props.onValueChange) props.onValueChange(value);
```
dhruvdutt: Not completely sure about this one.

packages/native/src/components/basic/TextField.js
```diff
@@ -83,19 +83,20 @@ class TextField extends Component {
 		const performUpdate = () => {
 			this.setState({
 				currentValue: value,
+			}, () => {
```
dhruvdutt: Did we miss the `setState` callback earlier?

packages/maps/src/components/basic/GeoDistanceDropdown.js
```diff
@@ -213,6 +213,7 @@ class GeoDistanceDropdown extends Component {
 						this.updateQuery(this.state.currentDistance);
 					}
 					this.locked = false;
+					if (props.onValueChange) props.onValueChange(this.state.currentDistance);
```
metagrover: `onValueChange` should only be called when `this.state.currentDistance` is truthy. So, you should move this statement inside the above `if` block.

packages/maps/src/components/basic/GeoDistanceSlider.js
```diff
@@ -209,6 +209,7 @@ class GeoDistanceSlider extends Component {
 					if (this.state.currentDistance) {
 						this.updateQuery(this.state.currentDistance);
 					}
+					if (props.onValueChange) props.onValueChange(this.state.currentDistance);
```
metagrover: same as above

packages/maps/src/components/basic/GeoDistanceDropdown.js
metagrover: Also, we need to pass this: as the value in `onValueChange`

packages/native/src/components/basic/TextField.js
metagrover: yep, good find!

metagrover: But the code should be:

```
const performUpdate = () => {
	this.setState({
		currentValue: value,
	}, () => {
		if (isDefaultValue) {
			this.updateQuery(value, props);
			// [handle onValueChange here]
		} else {
			// debounce for handling text while typing
			this.handleTextChange(value); // [handle onValueChange inside handleTextChange]
		}
		this.locked = false;
	});
};
```

packages/native/src/components/date/DateRange.js
metagrover: It seems correct. The pattern should always be:

1. state update
2. then, call updateQuery
3. call onValueChange

packages/native/src/components/range/RangeSlider.js
```diff
@@ -172,8 +172,10 @@ class RangeSlider extends Component {
 			this.setState({
 				currentValue,
 			}, () => {
-				this.updateQuery([currentValue[0], currentValue[1]], props);
+				const rangeValues = [currentValue[0], currentValue[1]];
+				this.updateQuery(rangeValues, props);
 				this.locked = false;
+				if (props.onValueChange) props.onValueChange(rangeValues);
```
metagrover: the format for the onValueChange's param here is wrong:

```
{
	start: currentValue[0],
	end: currentValue[1],
}
```

For determining the pattern, please check the second param in `checkValueChange`. Also, it would be great if you can fix this for web as well.

PR #333, opened 2018-04-17
maps: GeoDistanceDropdown - locationField support

### LINKED ISSUE #306: Feature: ability to specify dataField in GeoDistance components
**Issue Type:**

Enhancement

**Platform:**

Web / ReactiveMaps

**Description:**

Currently, the GeoDistance (Dropdown and Slider) components work by using Google's location search functionality, and then fire a query against the selected distance range.

This issue suggests that if a user already has location data in their ES index, we should allow them to optionally search against it (v/s strictly using Google location search).

This can be done by the use of `dataField` - when a user specifies the data field, they declare an intent of using this field to search the location data against (and use the ES index). When they don't specify this, Google's location search will be used against.

We can also take the location field ( (lat,lon) points) via `locationField` (currently, this is specified by dataField), which will eventually be used for firing the GeoQuery.

dhruvdutt:
Fixes #306

### DIFF
--- a/packages/maps/src/components/basic/GeoDistanceDropdown.js
+++ b/packages/maps/src/components/basic/GeoDistanceDropdown.js
@@ -14,8 +14,9 @@ import {
 	checkValueChange,
 	checkPropChange,
 	getClassName,
+	pushToAndClause,
 } from '@appbaseio/reactivecore/lib/utils/helper';
-
+import getSuggestions from '@appbaseio/reactivecore/lib/utils/suggestions';
 import types from '@appbaseio/reactivecore/lib/utils/types';
 
 import Title from '@appbaseio/reactivesearch/lib/styles/Title';
@@ -41,6 +42,7 @@ class GeoDistanceDropdown extends Component {
 			isOpen: false,
 		};
 		this.type = 'geo_distance';
+		this.internalComponent = `${props.componentId}__internal`;
 		this.locked = false;
 		this.coordinates = null;
 		this.autocompleteService = null;
@@ -53,6 +55,8 @@ class GeoDistanceDropdown extends Component {
 
 	componentWillMount() {
 		this.props.addComponent(this.props.componentId);
+		this.props.addComponent(this.internalComponent);
+
 		this.setReact(this.props);
 
 		if (this.props.selectedValue) {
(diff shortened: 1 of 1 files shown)


packages/maps/src/components/basic/GeoDistanceDropdown.js
```diff
@@ -445,6 +568,7 @@ GeoDistanceDropdown.propTypes = {
 	iconPosition: types.iconPosition,
 	innerClass: types.style,
 	innerRef: types.func,
+	locationField: types.stringOrArray,
```
metagrover: fuzziness and queryFormat props are missing in the prop-types.

(Is the linting check not woking properly?)

dhruvdutt: It doesn't work when we reassign `this.props` to `props` or something. Does it work for you? :thinking:

metagrover: oh yeah, it won't 👍

packages/maps/src/components/basic/GeoDistanceDropdown.js
```diff
@@ -589,12 +591,13 @@ GeoDistanceDropdown.propTypes = {
 };
 
 GeoDistanceDropdown.defaultProps = {
+	autoLocation: true,
 	className: null,
 	placeholder: 'Select a value',
 	showFilter: true,
 	style: {},
+	suggestions: [],
```
metagrover: This won't help because `suggestions: state.hits[props.componentId] && state.hits[props.componentId].hits` will overwrite suggestions value to `undefined` when hits are not available. It'd be best if you assign the defaultValue in the `mapStateToProps` function.

packages/maps/src/components/basic/GeoDistanceDropdown.js
```diff
@@ -93,6 +97,14 @@ class GeoDistanceDropdown extends Component {
 			this.updateQuery(this.state.currentDistance, nextProps);
 		});
 
+		if (Array.isArray(nextProps.suggestions)) {
```
metagrover: We should only consider `nextProps.suggestions` when `locationField` is set

packages/maps/src/components/basic/GeoDistanceDropdown.js
```diff
@@ -122,13 +134,107 @@ class GeoDistanceDropdown extends Component {
 
 	componentWillUnmount() {
 		this.props.removeComponent(this.props.componentId);
+		this.props.removeComponent(this.internalComponent);
 	}
 
-	setReact(props) {
-		if (props.react) {
-			props.watchComponent(props.componentId, props.react);
+	static suggestionsQuery = (value, props) => {
```
metagrover: This function should be renamed to `defaultQuery`. We've made it static for SSR reasons. Ideally, it should be identical to that of DataSearch otherwise SSR will cause serious problems

packages/maps/src/components/basic/GeoDistanceDropdown.js
dhruvdutt: Updated!

lakhansamani:
Fixed here #996

PR #367, opened 2018-05-15
web: Add onValueSelected for search components #254

### LINKED ISSUE #254: Add onValueSelected prop for search components
**Issue Type:**

Enhancement

**Description:**

Add a new prop `onValueSelected` to search components `DataSearch` and `CategorySearch`. This will accept a function which will be invoked with the current value when the selected value of the component changes (user presses enter or selects a suggestion or clicks outside). In case of `autosuggest` set to `false` this behaves like `onValueChange`. 

This prop will eliminate the need of having an extra component watching the search component with `react` prop. This also solves a use case of only using the search component for suggestions and displaying the results in a different page with the help of `URLParams`.

Update - We can also extend the `CategorySearch` by passing the selected category along with the value to this function. A very interesting and useful usecase is using this category to set the value of a different component, say `MultiList` as in the Car Store example app.

divyanshu013:
Resolves #254 

- introduces a new prop onValueSelected for `DataSearch` and `CategorySearch`
- changes blur behavior, does not auto blur after a selection
  has been made
- fixes an issue with stale values in onValueChange when selecting
  a suggestion
- fixes an issue with suggestions not displaying sometimes when
  clearing characters with backspace

If the changes look good, I can update the docs

### DIFF
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -359,7 +359,7 @@ class CategorySearch extends Component {
 
 	clearValue = () => {
 		this.setValue('', true);
-		this.onValueSelected('');
+		this.onValueSelected(null);
 	};
 
 	// only works if there's a change in downshift's value
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -359,17 +359,20 @@ class CategorySearch extends Component {
 
 	clearValue = () => {
 		this.setValue('', true);
+		this.onValueSelected('');
 	};
 
 	// only works if there's a change in downshift's value
 	handleOuterClick = () => {
 		this.setValue(this.state.currentValue, true);
+		this.onValueSelected();
 	};
 
-	handleKeyDown = (event) => {
-		if (event.key === 'Enter') {
-			event.target.blur();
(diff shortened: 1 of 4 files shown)


packages/web/src/components/search/CategorySearch.js
```diff
@@ -359,17 +359,20 @@ class CategorySearch extends Component {
 
 	clearValue = () => {
 		this.setValue('', true);
+		this.onValueSelected('');
```
metagrover: shouldn't it be `null`?

divyanshu013: Yeah, I'll update it to null

packages/web/src/components/search/CategorySearch.js
```diff
@@ -359,17 +359,20 @@ class CategorySearch extends Component {
...
 	};
 
 	// only works if there's a change in downshift's value
 	handleOuterClick = () => {
 		this.setValue(this.state.currentValue, true);
+		this.onValueSelected();
```
metagrover: Why are we not passing any value here? Shouldn't we be passing the `currentValue` here as well?

divyanshu013: The function's parameter defaults to `this.state.currentValue`.

metagrover: cool beans 👍

Issue #371, opened 2018-05-18
Infinite scroll with multi-page result and sorting are not working properly 


a-magdy:
Hi,

As always, I have to start with thanking you for your great work here :)

Issue Type: Bug

Platform: Web

Description: Sorting a multi-page result with infinite scroll breaks the component

Screenshots:
![screencapture-localhost-3001-2018-05-17-15_47_06](https://user-images.githubusercontent.com/14139855/40209889-52a3a650-5a42-11e8-973b-2dbf00195ba2.png)
<img width="961" alt="skaermbillede 2018-05-17 kl 15 43 39" src="https://user-images.githubusercontent.com/14139855/40209900-5e54c5b0-5a42-11e8-8b56-0fc31558af2b.png">

Minimal reproduction of the problem with instructions:
- From the appbaseio-app/booksearch demo, update the pagination to `false`
- Add some filter so that the results are easier to trace by eye, I choose to filter by author: `Michael Connelly (28)`
- Load more than 1 page using infinite loader
- Change the sorting
- Notice that the _msearch request after changing the sort is still using the same page number as the last page from before the sort, if the last request before changing the sort was: 
```
{"preference":"results"}
{"query":{"bool":{"must":[{"bool":{"must":[{"range":{"original_publication_year":{"gte":1970,"lte":2017,"boost":2}}},{"terms":{"authors.raw":["Michael Connelly"]}}]}}]}},"size":10,"from":20,"sort":[{"average_rating":{"order":"desc"}}]}
```
Notice the `from: 20`
The first request after changing the sort will be:
```
{"preference":"results"}
{"query":{"bool":{"must":[{"bool":{"must":[{"range":{"original_publication_year":{"gte":1970,"lte":2017,"boost":2}}},{"terms":{"authors.raw":["Michael Connelly"]}}]}}]}},"size":10,"from":20,"sort":[{"original_title.raw":{"order":"asc"}}]}
```
Notice the `from: 20` is still the same
- Now when you scroll to load another page, this is when you'll see duplicate results and console errors from react complaining about duplicate keys in the list
- This behaviour (saving the page number on sort change) works when pagination=true (not sure if it is the best approach though), but it breaks the infinite loader
[...]

PR #372, opened 2018-05-18
Fix issue #371 resetting page number on sort change

a-magdy:
Fix for #371 

- Update ReactiveList `handleSortChange`
  - Send the options.from = 0
  - setState({ from: 0, currentPage: 0 }) setting the correct page in the state, and force re-render the correct pagination with the first page selected


metagrover:
@a-magdy Thank you for contributing 🎉 

We will roll this out in the upcoming release.

PR #424, opened 2018-06-24
move-expo to dependency

vettedeng:
This break 0.8.0 build since expo is required in ReactiveMap


metagrover:
We use `expo` as a `peer-dependency`. Including this as a dependency can cause resolve module errors at build time.

Issue #421, opened 2018-06-22
DateRange dataField prop should accept both string and array


metagrover:
**Issue Type:**
Bug

**Platform:**
Web

**Description:**
We should accept both string and array in the [dataField's propTypes here](https://github.com/appbaseio/reactivesearch/blob/dev/packages/web/src/components/date/DateRange.js#L401).

Related: https://github.com/appbaseio/reactive-manual/issues/87

vivek12345:
@metagrover This https://github.com/appbaseio/reactivesearch/pull/426
Takes care of this change too

metagrover:
Thanks for contributing @vivek12345 💯 

We will roll this out in the upcoming release.

Issue #401, opened 2018-06-11
Add onFilterCleared prop to SelectedFilters


calebdel:
**Issue Type:**
Enhancement

**Platform:**
Web

**Description:**
Add a new prop `onFilterCleared` to `SelectedFilters` that will be invoked with the value of the selected filter being cleared or it will indicate that all filters have been cleared.

**Reactivesearch version:**
2.6.2

vivek12345:
@metagrover Any example with the code which renders filters?I have figured out how to do this, but I can't seem to figure out from the documentation as to how do I create a list with filters in it.

vivek12345:
Actually I figured out how to add filters.

metagrover:
Sure, we have the [components playground](https://opensource.appbase.io/playground/?selectedKind=Base%20components%2FDataController&selectedStory=Basic&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) as a submodule in this project that you can start with - which runs on the storybook, or alternatively, [web examples](https://github.com/appbaseio/reactivesearch/tree/dev/packages/web/examples) should come handy.

[SelectedFilters docs](https://opensource.appbase.io/reactive-manual/base-components/selectedfilters.html) is a good starting point. Here, we need to add support for `onFilterCleared` prop as mentioned above.

Thanks for taking out time to look into this 👍

vivek12345:
@metagrover should this be called also when someone unchecks the filter or only when I click on the cross icon of the filter to clear it?

metagrover:
I think the best thing to do here is to only call this function when there is an interaction with the `SelectedFilters` component, i.e. a filter is cleared. We can call this prop as `onClear`.

Also, in order to watch for the changes on the `SelectedFilters`, it'd be best to do that via `onChange` method, which will be executed anytime there is a change (addition or removal) in selected-values from the redux store.

vivek12345:
Alright that's a fair point.I am thinking how will I determine the interaction.Filters can be anything right.A dropdown, a checkbox or something else?

vivek12345:
Okay I think may be setValue can help here because that is called if something set or removed.What do you think?

metagrover:
Ideally, you should watch for the changes on the `selectedValues` from the redux store via `componentWillReceiveProps` method. But don't worry about that, we don't need to implement that right now - until a valid use-case pops up.

We should only focus on the `onClear` here for now 🐱

[...]

PR #426, opened 2018-06-25
onClear Prop added to SelectedFilters component and dataField prop should accept both string and array

vivek12345:
- Added a new prop onClear to SelectedFilters that will be invoked with the component name and the value of the selected filter being cleared.

- When clearAll is clicked this function will be called with just param and that would be null to indicate that the clearAll has been called.

- DateRange dataField prop should accept both string and array 

This is fix for https://github.com/appbaseio/reactivesearch/issues/401
and
https://github.com/appbaseio/reactivesearch/issues/421

### DIFF
--- a/packages/web/src/components/basic/SelectedFilters.js
+++ b/packages/web/src/components/basic/SelectedFilters.js
@@ -49,7 +49,7 @@ class SelectedFilters extends Component {
 
 							if (label && ((isArray && value.length) || (!isArray && value))) {
 								hasValues = true;
-								let valueToRender = this.renderValue(value, isArray);
+								const valueToRender = this.renderValue(value, isArray);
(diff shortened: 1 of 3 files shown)


packages/web/src/components/basic/SelectedFilters.js
```diff
@@ -43,14 +49,15 @@ class SelectedFilters extends Component {
 
 							if (label && ((isArray && value.length) || (!isArray && value))) {
 								hasValues = true;
+								let valueToRender = this.renderValue(value, isArray);
```
metagrover: It should be a `const`.

vivek12345: done

PR #428, opened 2018-06-26
Date Range Error

### LINKED ISSUE #151: Date Range component
## Bug Report

**Current behavior:**
- **Incorrect range selection**

If you have an end-date selected (screenshot - `2018-02-14`), and clear the start-date, you will be able to select any date in the calendar as the start-date, even the dates after `2018-02-14`.

screen shot 2017-12-27 at 8 51 31 pm

Also, if I have an end date selected, and I'm selecting the start date, the calendar should grey-out irrelevant dates (i.e. the dates after the end date)

- **UX issues**

    * Currently, the component does not tell you in any way which date are you selecting at any given point. (i.e. you won't know if you're selecting an end-date or a start-date)
    * The range selection isn't continuous. Calendar closes after selecting the start-date and you will have to click the end-date input field to open the calendar again to select the end date.

- **Support more Customization**

Allow users to change the middle `-`'s markup which connects two input fields together. Possible usecase: You may want to change the `-` to an arrow (`->`) or `to`.

**Expected / new behavior:**

- To improve the UX, we can highlight the range while selecting and keep the calendar open until the user is done selecting both the dates. 

**Reactivesearch version:**  latest

jyash97:
For #151

### DIFF
--- a/packages/web/src/components/date/DateRange.js
+++ b/packages/web/src/components/date/DateRange.js
@@ -30,6 +30,7 @@ class DateRange extends Component {
 		super(props);
 		this.state = {
 			currentDate: null,
+			dateHovered:null
 		};
 		this.locked = false;
 		props.setQueryListener(props.componentId, props.onQueryChange, null);
@@ -212,6 +213,12 @@ class DateRange extends Component {
 		});
 	}
 
+	handleDayMouseEnter = (day) => {
+		this.setState({
+			enteredTo: day,
+		});
+	}
+
 	handleDateChange = (
 		currentDate,
 		isDefaultValue = false,
@@ -266,13 +273,14 @@ class DateRange extends Component {
 	};
 
 	render() {
-		const { currentDate } = this.state;
+		const { currentDate,dateHovered } = this.state;
 		const start = currentDate ? currentDate.start : '';
 		const end = currentDate ? currentDate.end : '';
+		const dateHovered = dateHovered || '';
 		const selectedDays = [
-			start, { from: start, to: end },
+			start, { from: start, to: dateHovered },
 		];
-		const modifiers = { start, end };
+		const modifiers = { start, end:dateHovered };
 		return (
 			<DateContainer
 				range
@@ -353,6 +361,7 @@ class DateRange extends Component {
 							dayPickerProps={{
 								numberOfMonths: this.props.numberOfMonths,
 								initialMonth: this.props.initialMonth,
+								onDayMouseEnter: this.handleDayMouseEnter,
 								disabledDays: { before: this.state.currentDate ? this.state.currentDate.start : '' },
 								selectedDays,
 								modifiers,
--- a/packages/web/src/components/date/DateRange.js
+++ b/packages/web/src/components/date/DateRange.js
@@ -30,7 +30,7 @@ class DateRange extends Component {
 		super(props);
 		this.state = {
 			currentDate: null,
-			dateHovered:null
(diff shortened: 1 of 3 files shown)


divyanshu013:
Hi @jyash97, thanks for contributing 🤝

The UX issue is fixed with selecting date beyound the end date.

From the issue, here are the things which are missing before I can close this:

> Currently, the component does not tell you in any way which date are you selecting at any given point. (i.e. you won't know if you're selecting an end-date or a start-date)

For this you could implement the hover effect. [Example from docs](http://react-day-picker.js.org/examples/selected-range-enter)

> Allow users to change the middle -'s markup which connects two input fields together. Possible usecase: You may want to change the - to an arrow (->) or to.

> For custom rendering, we can use `onDayChange` from react-day-picker to expose the current selection and render as we please with a custom trigger to open the calendar view.

> When an end date is selected and the start date is cleared, we should still show the end date as marked partially.

Currently, it displays like the following:

image

You could implement a similar behavior which we have when only the start date is selected like:

image

(Here the starting date is partially selected to give a visual cue)

jyash97:
Highlighting the `End Date` :

screen shot 2018-06-27 at 11 10 53 am

And for the hovering effect, I think we will need to Change the methods, to onMouseEnter and onDayClick instead of onDayChange.

divyanshu013:
Yes, check out the example I shared (http://react-day-picker.js.org/examples/selected-range-enter) on how they're doing it. A similar approach should work here IMO, you might need some new variables in state to keep track of the hovered date range.

jyash97:
I tried it but the function mouse enter is not working.

divyanshu013:
Did you try this one? http://react-day-picker.js.org/api/DayPicker#onDayMouseEnter

jyash97:
Yes I tried this one and I checked by putting a console message on the first line and it didn't executed the message.

metagrover:
Can you try upgrading `react-day-picker` to the latest version and see if that helps?

metagrover:
Can you try upgrading `react-day-picker` to the latest version and see if that helps?

Also, we need `onDayMouseEnter` on `DayPicker` (not `DayPickerInput`) for hover to work.

jyash97:
In the component `DayPickerInput ` is used so may I change it to `DayPicker`?

divyanshu013:
You could use `dayPickerProps` to pass props to the `DayPicker` rendered by `DayPickerInput` http://react-day-picker.js.org/api/DayPickerInput#dayPickerProps

divyanshu013:
@jyash97 the hover effect isn't working, rest of the things look good. Also there are linting errors in the file.

packages/web/src/components/date/DateRange.js
```diff
@@ -212,6 +213,12 @@ class DateRange extends Component {
 		});
 	}
 
+	handleDayMouseEnter = (day) => {
+		this.setState({
+			enteredTo: day,
```
divyanshu013: This variable isn't being used, did you mean to use `dateHovered`?

jyash97: Yes sorry, I forgot to change this name, it should be `dateHovered` and everything would then be working

divyanshu013: I tested but it wasn't could you test it out first?

jyash97:
screen1

divyanshu013:
Thanks for contributing @jyash97 ✌️

Issue #394, opened 2018-06-08
Cursor jumps to end of input when inputValue is changed when using DataSearch


Arvind6353:
Hi , 
I am using Datasearch Component . 
Whenever I try to modify the search value (lets say change the middle character) , the cursor position is going to the end of the input . I guess its rerendering every time . 

Pls suggest a fix . 

I tried the demo sandbox example for datasearch and I face the same issue there .

` <DataSearch
            componentId="searchbox"
            dataField={["title"]}
            fieldWeights={[3]}
            placeholder="Search"
            autosuggest={false}
            highlight={true} />`

Arvind6353:
@divyanshu013  Pls do let me know if you find any soln or workaround to solve this issue

Arvind6353:
@metagrover @siddharthlatest : Do you have any soln or workaround for this issue ?

metagrover:
Sorry, didn't get any time to dive into this 😅

Open to PRs if you can propose a solution to this 🐱

Arvind6353:
@metagrover : I used ReactiveComponent and created a custom input and it solved this issue . 
But the Reactive component provides only an option to add the query . I am not able to find an option to highlight the fields. Can you pls let me know if there is an option for highlighting the results , similar to using setQuery in reactive component.

Arvind6353:
I solved this cursor issue by making some changes in Datasearch component file . I removed some state changes on the field (isOpen) and as I was not using suggestions here , I commented some of the state changes pertaining to it and it worked.

Arvind6353:
@aeronesto : Just comment out the state changes on isOpen field . And if you are not using suggestions , remove the state change on suggestions field as well .

aeronesto:
@Arvind6353  : I tried to do what you said, with combinations of autosuggest & this.state.isOpen set to true/false, but it didn't solve the cursor issue.

I am using @appbaseio/reactivesearch@2.6.2

divyanshu013:
@aeronesto, would you like to send a PR for this?

aeronesto:
@divyanshu013, sure thing.

rohit-sourcefuse:
@aeronesto, I am still facing this issue is it resolved?
Packages versions :
 "@appbaseio/reactivesearch": "^2.8.1"
"react": "16.4.1"

divyanshu013:
Hi @rohit-sourcefuse, it isn't resolved yet.

rohit-sourcefuse:
[...]

PR #429, opened 2018-06-26
Fixed cursor jumping to end of input after every input change

aeronesto:
Fixes: [#394](https://github.com/appbaseio/reactivesearch/issues/394)


aeronesto:
Created other bugs: clicking on a suggestion does not update the input field, and spaces are trimmed as you type.

PR #451, opened 2018-07-16
web: check for react prop changes in ReactiveComponent

Iaaan:
This checks for changes to `ReactiveComponent`'s `react` prop after the component has mounted in order to update the state accordingly - similar to several of the other web UI components.


metagrover:
Thanks for contributing @Iaaan 🎉

Iaaan:
👍 👍 Thanks @metagrover !

PR #464, opened 2018-08-01
Added include & exclude Fields as query options in Result Components

bietkul:


packages/web/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 <details>
     <summary>The following changes have been included in the <code>dev</code> branch and will be out in the next release. <b>Click to expand</b></summary>
+    - Added include & exclude Fields as query options in Result Components = [#464](https://github.com/appbaseio/reactivesearch/pull/464)
```
divyanshu013: Lets reference the issue here.

packages/web/src/components/result/ReactiveList.js
```diff
@@ -536,6 +538,7 @@ ReactiveList.propTypes = {
 	pagination: types.bool,
 	paginationAt: types.paginationAt,
 	react: types.react,
+	renderNoResults: types.func,
```
metagrover: why is this here? @bietkul 

It should be in a different PR.

Issue #465, opened 2018-08-01
maps: GeoDistanceDropdown Allow to set Google maps Restrictions by countries


durdenx:
**Enhancement:**
Currently autocompletion is global, allow to set Google maps Restriction to limit places autocompletion by countries for `GeoDistanceSlider`and `GeoDistanceDropdown`.

In gmaps doc: `autocomplete.setComponentRestrictions({'country': countries});`
see https://developers.google.com/maps/documentation/javascript/examples/places-autocomplete-multiple-countries

**Platform:**
Web

metagrover:
This is pretty neat. Open to accepting PRs on this, if you or anyone wants to take this up.

metagrover:
This will be released in the upcoming release of maps (~ this week).

Props to @jyash97 for working on this 💯

metagrover:
Released in `v2.0.0` 🎉

PR #475, opened 2018-08-13
added country  restriction props in Geodropdown

jyash97:
Solves #465


packages/maps/src/components/basic/GeoDistanceSlider.js
```diff
@@ -456,6 +461,7 @@ GeoDistanceSlider.propTypes = {
 	beforeValueChange: types.func,
 	className: types.string,
 	componentId: types.stringRequired,
+	countries: types.stringOrArray,
```
metagrover: It should always be an array of strings

Issue #441, opened 2018-07-11
enable pagination during SSR


aol-nnov:
**Issue Type:** enhancement

**Platform:** Web

**Description:**

For me, the main idea behind SSR is to make web app crawlable (including dummy crawlers, which do not execute javascript).

So, for testing purposes I fired your demo `packages/web/examples/ssr`, disabled javascript in the browser and navigated to `localhost:3000`.

First page renders just fine, but pagination buttons do not function at all until I enable js back.
It would be nice, if pagination buttons work without js and with `URLParams` enabled in `ResultCard`

**Reactivesearch version:** dev branch

**Browser:** all

metagrover:
This is interesting.

I think if we add the `href` attribute on the pagination `<a>` buttons and add `event.preventDefault()` in the `onClick` function definition, it can work.

Do you want to give this a shot? Open to accepting PRs on this.

aol-nnov:
@metagrover I've just tried to add `?SearchResult-page=2` while js is disabled and nothing happened. Seems like I need the same answer  as in #440 :)

metagrover:
`?SearchResult-page=2` is correct. It takes the query param from the request object.

It seems I missed out on the implementation for considering `pages` in SSR. I will check and get back to you on this 🐱

metagrover:
@aol-nnov Thanks for the patience on this. I will look into it today.

metagrover:
Fixed here: c6d3be2 and will be released in `2.7.0` 🎉

aol-nnov:
Deepk, unfortunately I still do not see any changes in 2.7.0

Here is how I perform test: 
the very same `packages/web/examples/ssr` example from `dev` branch with minor change:
```patch
diff --git a/packages/web/examples/ssr/pages/index.js b/packages/web/examples/ssr/pages/index.js
index f2e49a4..6d6d6df 100644
--- a/packages/web/examples/ssr/pages/index.js
+++ b/packages/web/examples/ssr/pages/index.js
@@ -78,6 +78,7 @@ const components = {
                        url: data.listing_url,
                }),
                pagination: true,
+              URLParams: true,
                react: {
                        and: ['SearchSensor', 'GuestSensor'],
                },
```

then `npm install && npm run dev`
And finally I launch firefox with `javascript.enabled: false` in `about:config`

[...]

PR #476, opened 2018-08-14
Get initial search state from query params to make SSR pagination work.

aol-nnov:
a followup for discussion in #441 

pagination in this example now works even with js disabled, but for now you'll need to modify url manually (i.e. `?SearchResult=1` or `?SearchResult=2`) because pagination component needs some changes to be clickable while js is disabled.


metagrover:
Awesome, thanks! 💯

Issue #420, opened 2018-06-22
web: Replace shade function


divyanshu013:
**Issue Type:**

Enhancement

**Platform:**

Web

**Description:**

The current `shade` function doesn't provide expected shades of colors correctly. An alternative is to switch to `polished` or some other library which gives better results for all `primaryColor` theming.

**Reactivesearch version:** x.y.z

2.6.8

PR #482, opened 2018-08-15
web: Add tabIndex in pagination buttons and enter key event to change…

### LINKED ISSUE #448: Pagination in reactivesearch using ResultList component and 508 compliance
We are using pagination in search result. We use ResultList component. Additional to click on page 2, 3 and to support 508 compliance, we need to use tab in order to go to different pages .
How can we do it?

jyash97:
fixes #448

### DIFF
--- a/packages/web/src/components/result/addons/Pagination.js
+++ b/packages/web/src/components/result/addons/Pagination.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { getClassName } from '@appbaseio/reactivecore/lib/utils/helper';
+import { getClassName, handleA11yAction } from '@appbaseio/reactivecore/lib/utils/helper';
 import types from '@appbaseio/reactivecore/lib/utils/types';
 
 import Button, { pagination } from '../../../styles/Button';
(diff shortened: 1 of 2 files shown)


packages/web/src/components/result/addons/Pagination.js
```diff
@@ -42,6 +42,8 @@ export default function Pagination(props) {
 					className={className}
 					primary={primary}
 					key={i - 1}
+					tabIndex="0"
+					onKeyPress={event => (event.key === 'Enter' ? props.setPage(i - 1) : null)}
```
metagrover: Looks good, but we already have [this helper method](https://github.com/appbaseio/reactivecore/blob/master/src/utils/helper.js#L228-L233) that you can use here.

divyanshu013:
Requires #420 to get correct shade of primary color.

divyanshu013:
Merged [here](https://github.com/appbaseio/reactivesearch/commit/dd4bf9eeb0644ec7d639c2946bd3b55e778f2b98)

packages/web/src/components/result/addons/Pagination.js
```diff
@@ -42,6 +42,8 @@ export default function Pagination(props) {
 					className={className}
 					primary={primary}
 					key={i - 1}
+					tabIndex="0"
+					onKeyPress={event => handleA11yAction(event, props.setPage(i - 1))}
```
divyanshu013: When passing a callback, it should be a function so the execution can be delayed. For example here the second parameter should have been a function i.e. `() => props.setPage(...)`

PR #525, opened 2018-09-17
web: add analytics support

### LINKED ISSUE #484: Feature: Add support for recording analytics
**Issue Type:**

Enhancement

**Platform:**

ReactiveSearch for web / mobile

**Description:**

Add a boolean `analytics` prop in ReactiveBase component which when set to true, allows recording search analytics (and click analytics) when appbase.io is used as a backend.

Since the Elasticsearch Query DSL is huge and offers a variety of ways to search, we require setting explicit headers to describe the search intent. These headers can be set by the ReactiveSearch library in the outgoing requests when the user had opted-in for recording analytics.

Recording analytics is based on the following headers:

`X-Search-Query` -> value represents the search query term (the input value in DataSearch, CategorySearch)

`X-Search-Id` -> An existing Id (aka search session) on which to apply the search query. This is returned as a response header by the backend for each search query. (In ReactiveSearch, only the active query's Id needs to be maintained. When a user applies a consequent filter, clicks on a result, etc., we want to attribute the changes to the same search session)

`X-Search-Filters` -> value is of type: "key1=value1,key2=value2,..." where key represents the filter component and value represents the selected value. (If the same filter has multiple values selected, they should be passed as "key1=value1,key1=value2,...")

---

`X-Search-Click` -> value is of type true / false,

`X-Search-ClickPosition` -> value is of type Number (e.g. 1, 2 denoting the result item being clicked)

These last three events allow recording click events and conversion events, a click event should be triggered when an end-user clicks on one of the search results. Since these events record what happens after a search query is fired, they should also be accompanied with a `X-Search-Id` header.

---

To successfully record a search analytics event, at least one of `X-Search-Query` (denoting a new query) or `X-Search-Id` header (denoting an existing search query) should be present.

2.7.0

**Anything else:**

divyanshu013:
- ReactiveCore would use these to set the analytics related values

Resolves #484. This PR is necessary for the analytics logic in reactivecore to work. In the next PR I'll include support for filters and click analytics.

[Docs](https://opensource.appbase.io/reactive-manual/advanced/analytics.html)

### DIFF
--- a/packages/web/CHANGELOG.md
+++ b/packages/web/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Dynamically change include/exclude Fields [#417](https://github.com/appbaseio/reactivesearch/issues/417)
 - Fix empty categories in suggestions for `CategorySearch` [#502](https://github.com/appbaseio/reactivesearch/issues/502)
 - Fix `defaultSuggestions` not appearing in `CategorySearch` [commit](https://github.com/appbaseio/reactivesearch/commit/776074f766d5e4c14759ed36e1531d234df35e91)
+- Add analytics support [#484](https://github.com/appbaseio/reactivesearch/issues/484)
 
 </details>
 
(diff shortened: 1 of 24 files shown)


divyanshu013:
This PR now includes all the features for analytics recording. Its also dependent on the [corresponding PR](https://github.com/appbaseio/reactivecore/pull/15) in reactivecore.

Issue #523, opened 2018-09-17
showFilter not present in Typescript Definitions for v2.8.2 on MultiList


mb-3000:
**Issue Type:**
bug

**Platform:**
Web

**Description:**
The docs state Multilist etc have a parameter on them `showList`. 

https://opensource.appbase.io/reactive-manual/list-components/multilist.html

This is not present in the interfaces defined in Typescript:

**Code:**
```
import * as React from 'react';
import { CommonProps } from '../../';
import * as types from '../../types';

export interface MultiList extends CommonProps {
	beforeValueChange?: (...args: any[]) => any;
	customQuery?: (...args: any[]) => any;
	dataField: string;
	defaultSelected?: types.stringArray;
	filterLabel?: string;
	innerClass?: types.style;
	onValueChange?: (...args: any[]) => any;
	placeholder?: string;
	queryFormat?: types.queryFormatSearch;
	react?: types.react;
	renderListItem?: (...args: any[]) => any;
	selectAllLabel?: string;
	showCheckbox: boolean;
	showCount?: boolean;
	showSearch?: boolean;
	size?: number;
	sortBy?: types.sortByWithCount;
	themePreset?: types.themePreset;
	title?: types.title;
	showMissing?: boolean;
	missingLabel?: string;
}

declare const MultiList: React.ComponentType<MultiList>;

export default MultiList;

```

In a few of the other list objects this param seems to be there in and then others it isn't but it's noted in the docs that it should be.

**Minimal reproduction of the problem with instructions:**
Try to write a component with showFilter={true} like the docs state and you will get a TS compiler error or marker in your IDE if you are using IDEA etc.

**Reactivesearch version:** 2.8.2

Not sure if this a docs issue or perhaps an issue with the typescript binding. Apologies for my confusion, otherwise a great set of components!

If I add the line in manually to the d.ts file it works.

```
import * as React from 'react';
import { CommonProps } from '../../';
import * as types from '../../types';

export interface MultiList extends CommonProps {
	beforeValueChange?: (...args: any[]) => any;
	customQuery?: (...args: any[]) => any;
	dataField: string;
	defaultSelected?: types.stringArray;
	filterLabel?: string;
	innerClass?: types.style;
	onValueChange?: (...args: any[]) => any;
	placeholder?: string;
	queryFormat?: types.queryFormatSearch;
	react?: types.react;
	renderListItem?: (...args: any[]) => any;
	selectAllLabel?: string;
	showCheckbox: boolean;
[...]

PR #526, opened 2018-09-17
fix to mappings file for multilist

mb-3000:
I made the fix as we discussed in https://github.com/appbaseio/reactivesearch/issues/523

I also had to add a few things to the package.json to get things to run. eslint also throws a lot of errors but that might be just CRLF/LF issues on my mac here.

Change is minimal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,10 +25,13 @@
 		]
 	},
 	"devDependencies": {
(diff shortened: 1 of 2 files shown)


metagrover:
@mb-3000 Thanks for contributing 💯

mb-3000:
your welcome.

PR #527, opened 2018-09-18
Result of defaultSelected function not evaluated by isEqual

bietkul:


packages/web/src/components/range/DynamicRangeSlider.js
```diff
@@ -77,7 +77,7 @@ class DynamicRangeSlider extends Component {
 					Math.ceil(nextProps.range.end),
 				], nextProps);
 			}
-		} else if (!isEqual(this.props.defaultSelected, nextProps.defaultSelected) && nextProps.range) {
+		} else if ((this.props.defaultSelected !== nextProps.defaultSelected) && nextProps.range) {
```
metagrover: `defaultSelected` prop is a function here, which can be an arrow function as well. We don't want this change to get triggered unnecessarily in case of re-renders. Hence, ideally we should check for the result of the function call i.e. `defaultSelected(value)` and compare the results of the two function calls to check for equality.

PR #530, opened 2018-09-19
web: Improve suggestion rendering in search components

### LINKED ISSUE #507: Feature: Improve UI/Ux of suggestions for DataSearch and CategorySearch
**Issue Type:**

Enhancement

**Platform:**

ReactiveSearch for Web

**Description:**

Incorporate different UI/Ux improvements to make the suggestions UI more palatable:
1. Use more symmetry in how the suggestions are displayed,
2. Add subtle highlighting,
3. Check for spacing and padding.

**Screenshots:**

**Anything else:**

### LINKED ISSUE #524: web: Filter suggestions which are passed via onSuggestion prop
**Issue Type:**

Enhancement/Bug

**Platform:**

Web

**Description:**

When using `onSuggestion` prop we currently pass all the suggestions. But since it's a render prop per suggestion, it's tricky to implement a custom parsing logic by the user. Also these suggestions may include duplicates which are therefore difficult to remove. We should send only the parsed suggestions in this prop. If someone is interested in using unfiltered suggestions, the `renderSuggestions` prop is more useful.

**Screenshots:**

https://imgur.com/a/tUte6nm

Use `onSuggestion` prop on a search component.

2.8.2

divyanshu013:
Resolves #507 and fixes #524 with `renderSuggestion` prop

- Adds highlighting
- Adds `renderSuggestion` prop
- Deprecates `onSuggestion` prop

### `renderSuggestion` usage

Similar to the existing `onSuggestion` prop usage but supports extra keys for `title`, `image` and `description`. Both title and description support highlighting. In order to take control of rendering specify a `label` key just like `onSuggestion` supports (this has the highest precedence while rendering).

```js
<DataSearch
  ...
  renderSuggestion={(suggestion) => ({
    title: suggestion.source.original_title,
    description: suggestion.source.authors,
    image: suggestion.source.image,
    value: suggestion.source.original_title,  // required
    // optionally render the entire JSX using label
    label: <JSX>,  // has higher precedence over title, description, image
    source: suggestion.source  // for onValueSelected to work with onSuggestion
  })}
/>
```

Also with `renderSuggestion` we don't/shouldn't implement the `url` part as specified in #507 since it introduces more complexity taking into consideration clicks or enter key presses. A better way to handle this is after a suggestion is selected using `onValueSelected`.

However if someone wishes to open links on clicking suggestions that should also work when using the `label` key in `renderSuggestion` and adding an `onClick` handler at the root.

It's also possible to select another value other than the rendered suggestion by specifying it in the `value` key. For example if someone wishes to only set a truncated value of max 20 characters they could do something like:

```js
{
  ...
  value: suggestion.source.original_title.slice(0, 20)
}
```

Docs can be updated later with the above usage when merged.

### DIFF
--- a/packages/web/src/components/search/DataSearch.js
+++ b/packages/web/src/components/search/DataSearch.js
@@ -29,7 +29,7 @@ import CancelSvg from '../shared/CancelSvg';
 import InputIcon from '../../styles/InputIcon';
 import Container from '../../styles/Container';
 import { connect } from '../../utils';
-
+import SuggestionItem from './addons/SuggestionItem';
 
 class DataSearch extends Component {
 	constructor(props) {
@@ -238,18 +238,28 @@ class DataSearch extends Component {
 	};
 
 	onSuggestions = (results) => {
-		if (this.props.onSuggestion) {
-			return results.map(suggestion => this.props.onSuggestion(suggestion));
+		const { onSuggestion, renderSuggestion } = this.props;
+		if (onSuggestion) {
+			if (process.env.NODE_ENV !== 'production') {
+				console.warn('onSuggestion prop is deprecated please use renderSuggestion');
+			}
+			return results.map(suggestion => onSuggestion(suggestion));
 		}
 
 		const fields = Array.isArray(this.props.dataField)
 			? this.props.dataField : [this.props.dataField];
 
-		return getSuggestions(
+		const parsedSuggestions = getSuggestions(
 			fields,
 			results,
 			this.state.currentValue.toLowerCase(),
 		);
+
+		if (renderSuggestion) {
+			return parsedSuggestions.map(suggestion => renderSuggestion(suggestion));
+		}
+
+		return parsedSuggestions;
 	};
 
 	setValue = (value, isDefaultValue = false, props = this.props, cause) => {
@@ -438,6 +448,7 @@ class DataSearch extends Component {
 	);
 
 	render() {
+		const { currentValue } = this.state;
 		let suggestionsList = [];
 
 		if (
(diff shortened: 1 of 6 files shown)


packages/web/src/components/search/DataSearch.js
```diff
@@ -238,18 +238,28 @@ class DataSearch extends Component {
...
 
 	onSuggestions = (results) => {
-		if (this.props.onSuggestion) {
-			return results.map(suggestion => this.props.onSuggestion(suggestion));
+		const { onSuggestion, renderSuggestion } = this.props;
+		if (onSuggestion) {
+			if (process.env.NODE_ENV !== 'production') {
+				console.warn('onSuggestion prop is deprecated please use renderSuggestion');
+			}
+			return results.map(suggestion => onSuggestion(suggestion));
```
metagrover: Why are we not parsing suggestions for `onSuggestion`? Is it to avoid breaking the existing usage?

packages/web/src/components/search/addons/SuggestionItem.js
```diff
@@ -0,0 +1,77 @@
...
+			)
+			: label;
+	} else if (title || image || description) {
+		return (
+			<Flex alignItems="center" css={{ width: '100%' }}>
+				{
+					image
+					&& (
+						<div css={{ margin: 'auto', marginRight: 10 }}>
+							<img src={image} alt=" " height="50px" width="50px" css={{ objectFit: 'contain' }} />
```
metagrover: Would it better if we use `title` in `alt`?

packages/web/src/components/search/DataSearch.js
divyanshu013: Yeah, also since this is a breaking change we can use `renderSuggestion` which would receive parsed suggestions.

packages/web/src/components/search/addons/SuggestionItem.js
divyanshu013: Actually this was intentional. Otherwise, while the images load you will see the alt attribute in the dropdown which looks weird while the suggestions' images are loading. With `' '` the space shows up as blank which is then filled by the image.

PR #536, opened 2018-09-24
web: Add support for scrollOnChange

### LINKED ISSUE #528: jumping to top in rare cases
**Issue Type:**
bug

**Platform:**
web

**Description:**
When I have several filters I combine, at some point the page jumps to the very top. This not only happens when I receive 0 results but also when I still have results

see [sandbox](https://codesandbox.io/s/r53ro5735q). Click on `Nissan` and then `scroll down a little`. Then click on `5` (on the rating component)

**Reactivesearch version:**
2.8.0

**Browser:**
all

**Anything else:**

jyash97:
Fixes #528

### DIFF
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -210,7 +210,7 @@ class ReactiveList extends Component {
 						isLoading: false,
 					});
 
-					if (nextProps.hits.length < this.props.hits.length) {
+					if (nextProps.hits === this.props.hits) {
 						// query has changed
 						if (this.props.scrollOnChange) {
 							window.scrollTo(0, 0);
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -210,7 +210,7 @@ class ReactiveList extends Component {
 						isLoading: false,
 					});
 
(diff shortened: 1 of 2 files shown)


packages/web/src/components/result/ReactiveList.js
```diff
@@ -212,7 +212,9 @@ class ReactiveList extends Component {
 
 					if (nextProps.hits.length < this.props.hits.length) {
```
metagrover: I think @divyanshu013 and I discussed this in the related issue to change this `if` condition to have shallow equal comparison between the two `hits`.

packages/web/src/components/result/ReactiveList.js
```diff
@@ -210,7 +210,7 @@ class ReactiveList extends Component {
 						isLoading: false,
 					});
 
-					if (nextProps.hits.length < this.props.hits.length) {
+					if (nextProps.hits === this.props.hits) {
```
divyanshu013: This condition should be inversed, i.e. `nextProps.hits !== this.props.hits` because if the hits have changed only then the `if` block should execute.

metagrover:
@jyash97 can you change the base to `next` and make relevant changes (We use `componentDidUpdate` there).

jyash97:
Done in #729

PR #537, opened 2018-09-24
web: Add support for scrollTarget

### LINKED ISSUE #369: Infinite Scroll doesn't work if the page is not scrollable
Hi guys,

As always thanks for your great work in this library :)

Issue Type: Bug -an edge case with a workaround-

Platform: Web

Description: Infinite scroll doesn't work if the browser page is not scrollable

Screenshots:

- 10 items per page on a big screen:
skaermbillede 2018-05-17 kl 15 44 37

- or only 1 item per page (as an extreme case): 
skaermbillede 2018-05-17 kl 15 44 15

Minimal reproduction of the problem with instructions:

- From the booksearch demo, update the `pagination` to false, and update the page size to 10 or less until the scrollbar disappears 
- Try to scroll down to fetch more, but it won't be possible, until you resize the screen, so that the scrollbar appears, so that the infinite loader works

Reactivesearch version: 2.6.2

Browser: Chrome 66

I am not sure whether this is a real issue, but I thought to start a discussion about it, maybe we could come up with a solution for it in the future, like adding a load more button as proposed in this issue: https://github.com/appbaseio/reactivesearch/issues/126 in case the page is not scrollable, and we are not on the last page yet

Thanks :)

jyash97:
Fixes #369

### DIFF
--- a/packages/web/src/components/result/ReactiveList.d.ts
+++ b/packages/web/src/components/result/ReactiveList.d.ts
@@ -32,6 +32,7 @@ export interface ReactiveListProps extends CommonProps {
 	onPageClick?: (...args: any[]) => any;
 	defaultPage?: number;
 	listClass?: string;
+	scrollTarget?:string;
 }
 
 declare const ReactiveList: React.ComponentType<ReactiveListProps>;
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -110,7 +110,9 @@ class ReactiveList extends Component {
 		this.setReact(this.props);
 
 		if (!this.props.pagination) {
-			window.addEventListener('scroll', this.scrollHandler);
+			const { scrollTarget } = this.props;
+			this.domNode = scrollTarget ? document.getElementById(scrollTarget) : window;
+			this.domNode.addEventListener('scroll', this.scrollHandler);
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


GraemeFulton:
This is great, will it be merged? I also need this 👍

PierrePX:
i have no idea on how to do a pull request, but i did this on my end.
I have overriden the existing ResultList class

If someone's interested..

metagrover:
This will soon be merged (after some tests) and rolled out in the next release.

packages/web/src/components/result/ReactiveList.js
```diff
@@ -578,6 +588,7 @@ ReactiveList.propTypes = {
 	pagination: types.bool,
 	paginationAt: types.paginationAt,
 	react: types.react,
+	scrollTarget: types.string,
```
metagrover: same prop should be listed in other result components

packages/web/src/components/result/ReactiveList.d.ts
```diff
@@ -32,6 +32,7 @@ export interface ReactiveListProps extends CommonProps {
 	onPageClick?: (...args: any[]) => any;
 	defaultPage?: number;
 	listClass?: string;
+	scrollTarget?:string;
```
metagrover: same type should be listed in other result component types

metagrover:
Released in `v2.10.0` 🎉

PR #539, opened 2018-09-25
web: Open page 1 whenquery changes

jyash97:


metagrover:
I agree with @divyanshu013. Can you please cut down on the noise here and briefly explain the approach behind your solution?

jyash97:
Closing this because this created other issues

Issue #515, opened 2018-09-12
Non-limited search for MultiList with size


rap2hpoutre:
**Issue Type:**
Enhancement issue (or maybe just something I don't understand)

**Platform:**
ReactiveSearch for Web

**Description:**
Currently, the `MultiList` component displays a limited list that is useful for filtering. It's limited by `size`. There is also a "search" subcomponent associated to this component that can help to search inside the `MultiList` list if `showSearch` is true.

My problem is that it can only search into the limited list, not the whole list. So if I use the Multilist to filter the results of a `DataSearch`, I cannot search _inside_ the `MultiList` for words that are not displayed. So if I want to search for a word that is not in the first 100 results of the `MultiList`, it is not returned.

**Is there a way to search in the whole database when searching inside a `MultiList` rather than searching in the first 100 results?**

Example `MultiList` component (no really useful example IMO)
```
<MultiList
                    componentId="base"
                    dataField="BASE.keyword"
                    title="Domaine"
                    queryFormat="and"
                    filterLabel="Domaine "
                    className="filters"
                    showSearch={true}
                    showCheckbox={true}
                    URLParams={true}
                    placeholder="Rechercher"
                    react={{
                      and: FILTER
                    }}
                  />
```

**Reactivesearch version:** latest

**Browser:** all

siddharthlatest:
@rap2hpoutre The max size for a MultiList can be set to `1000` right now. We have an open issue where we working on adding pagination support to a List component (via `Show More`).

rap2hpoutre:
@siddharthlatest OK thank you!

> The max size for a MultiList can be set to 1000 right now
Uh, it's actually not an option because it will be slow (we have 5 Multilists on the same page).

> We have an open issue where we working on adding pagination support to a List component (via Show More).

Great news! Does it means that it will "lazy-load" the list on search/page-change (good) or load everything (not good for me)?

siddharthlatest:
> Great news! Does it means that it will "lazy-load" the list on search/page-change (good) or load everything (not good for us)?

It will lazy-load.

> My problem is that it can only search into the limited list, not the whole list.

I am not sure how (at least right now) we can address the issue you originally raised above. Any thoughts, @metagrover?

metagrover:
I do see value in it. In cases of large datasets, local search filtering won't be a great UX. Ideally, the search should happen on the ES in such cases and maybe we can keep it as an optional behavior via some boolean prop.

[...]

PR #542, opened 2018-09-27
loadMoreLabel

rap2hpoutre:
See: https://github.com/appbaseio/reactivesearch/issues/515#issuecomment-425150797

The prop is a `string` to be consistent with `missingLabel`, `filterLabel`, etc. It's called `*Label` for the same reason. Still, I can edit the PR and change behavior.

### DIFF
--- a/packages/web/src/components/list/MultiDropdownList.d.ts
+++ b/packages/web/src/components/list/MultiDropdownList.d.ts
@@ -25,6 +25,7 @@ export interface MultiDropdownList extends CommonProps {
 	showMissing?: boolean;
 	missingLabel?: string;
 	showLoadMore?: boolean;
+	loadMoreLabel?: types.title;
 }
 
 declare const MultiDropdownList: React.ComponentType<MultiDropdownList>;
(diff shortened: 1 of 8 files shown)


rap2hpoutre:
@metagrover Thank you for your review. 👍 

I'm not sure to understand: 

`title` prop is of type `title`. Do you want me to change the type of `loadMoreLabel` to `title`? (that sounds a bit strange to me, that's why I feel like I misunderstood something :) !).

I checked types here: https://github.com/appbaseio/reactivecore/blob/master/src/utils/types.js and saw that `title` is:

	title: oneOfType([string, any]),

Do you want me to create a new `label` type in another PR on this other repo before?

I'm a bit lost 😅

metagrover:
@rap2hpoutre You can use `title` type with `loadMoreLabel`.

rap2hpoutre:
@metagrover Done!

rap2hpoutre:
@metagrover Done! 🎉

rap2hpoutre:
@divyanshu013 @metagrover Done.

metagrover:
@rap2hpoutre Thanks for contributing 🤝

davidkong0987:
how are loadMore and handleLoadMore used in ReactiveList? They seem like interesting properties but I couldn't find much on them.

PR #553, opened 2018-10-04
Tooltip in RangeSlider

### LINKED ISSUE #284: In DynamicRange Silder Tooltip is not present in V2.0.0

jyash97:
`Rheostat` has method called `handle` to create custom Handle, which gets following values :
 aria-valuenow , aria-valuemax ,aria-valuemin, className, style and other methods.

Used this handle method which conditionally renders Component if user wants to have tooltip, he can use `handleTooltip` render prop:

```js
<RangeSlider
    handleTooltip = (children, value) => <Tooltip value={value} >{children}</Tooltip>
/>
```
If this render prop is not present we can render the default `Handle` component.
In this way we wont need to add `Tooltip` component in RS and user can use his own Tooltip.

## Issue
I tried doing the same was getting the dom element but it was not visible on UI.

Will close #284


jyash97:
New API : 
```js 
<RangeSlider 
  showTooltip 
/>
```

Tooltip visible on `active`, `hover` and `focus` state of handle.

tooltip_slider

If this looks good I can go on to add them in other Slider Components

davidklebanoff:
This is looking great! I do feel like from a usability perspective, there may be times when someone wants the tooltips to show even when not hover/focused. Especially for smaller ranges like 1-20. It may be worth adding a configuration, or at least some example code in the docs to allow such behavior.

divyanshu013:
That's a good point. Lets make this prop an `enum`. Lets have a single prop say, `tooltipTrigger`. It can have the following values:

- `focus` (only displays when in focus)
- `hover` (displays when hovered - this would also work in case of focus when mouse is used)
- `always` (always displays the tooltip)
- `none` (default - does not display tooltip. This also helps avoid any breaking changes in the API)

metagrover:
`tooltipTrigger` sounds good to me! 💯

jyash97:
Removed `showTooltip` instead introduced a new prop called `tooltipTrigger` as discussed above.
`tooltipTrigger` have values `none`, `hover`, `always` and `focus`. `none` by default.

---
 **ALWAYS**

tooltip_always

---
**FOCUS**

tooltip_focus

---
**HOVER**
tooltip_hover

---
**NONE**
tooltip_none

packages/maps/src/components/basic/GeoDistanceSlider.js
```diff
@@ -484,6 +495,7 @@ GeoDistanceSlider.propTypes = {
 	react: types.react,
 	showFilter: types.bool,
 	showIcon: types.bool,
+	tooltipTrigger: types.string,
```
metagrover: We should use `oneOf` prop-type here. You can add it to the reactivecore/types.

packages/web/src/components/range/DynamicRangeSlider.js
```diff
@@ -413,6 +424,7 @@ DynamicRangeSlider.propTypes = {
 	react: types.react,
 	showHistogram: types.bool,
 	showFilter: types.bool,
+	tooltipTrigger: types.string,
```
metagrover: update prop-type.

packages/web/src/components/range/RangeSlider.js
```diff
@@ -338,6 +349,7 @@ RangeSlider.propTypes = {
 	histogramQuery: types.func,
 	showFilter: types.bool,
 	showSlider: types.bool,
+	tooltipTrigger: types.string,
```
metagrover: update prop-type.

packages/web/src/styles/Slider.js
```diff
@@ -52,6 +52,48 @@ const Slider = styled('div')`
 		top: -10px;
 	}
 
+	.slider-tooltip,.slider-tooltip-focus,.slider-tooltip-visible {
```
metagrover: add proper spacing here

packages/web/src/styles/Slider.js
```diff
@@ -52,6 +52,48 @@ const Slider = styled('div')`
...
+    visibility: hidden;
+    background-color: black;
+    color: #fff;
+    text-align: center;
+    border-radius: 6px;
+    padding: 6px 8px;
+    position: absolute;
+    z-index: 1;
+		transform: translate(-50%,-170%); /* Positions Tooltip Container */
```
metagrover: linting issue

packages/web/src/styles/Slider.js
```diff
@@ -52,6 +52,48 @@ const Slider = styled('div')`
...
+	}
+
+	.slider-tooltip-visible {
+		visibility: visible;
+	}
+
+	.slider-tooltip:after,.slider-tooltip-focus:after,.slider-tooltip-visible:after{
+		content: '';
```
metagrover: linting issue + add proper spacing.

Issue #267, opened 2018-02-13
Add support for unrated results in RatingsFilter


davidklebanoff:
**Issue Type:**
enhancement

**Description:**
Sometimes search results may have no ratings. On the `RatingsFilter` component, it would be nice to have a boolean option that indicates whether unrated results should be included in a filter. The property could be set at the component level, but setting it at the `data` property level would allow more granularity.

The option could indicate whether to include a result when the dataField is `undefined` or 0.

Something like:

```jsx
<RatingsFilter
  componentId="ratingsSensor"
  dataField="ratings"
  data={[
    {"start": 4, "end": 5, "label": "4 & up"},
    {"start": 3, "end": 5, "label": "3 & up"},
    {"start": 1, "end": 5, includeUnrated: true, "label": "All"}
  ]}
/>
```

jyash97:
This would be included in the next release ✨ . Fixed here #1003

PR #558, opened 2018-10-05
web : Add includeUnrated in RatingsFilter

jyash97:
#267 

Getting a prop-type error in `SelectedValue`, rest everything is working 💯

### DIFF
--- a/packages/web/src/components/range/RatingsFilter.js
+++ b/packages/web/src/components/range/RatingsFilter.js
@@ -66,6 +66,7 @@ class RatingsFilter extends Component {
 					? [
 						nextProps.defaultSelected.start,
 						nextProps.defaultSelected.end,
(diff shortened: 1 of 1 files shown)


davidklebanoff:
Does this cover both possible cases, when the rating is `undefined` or `0`?

davidklebanoff:
> `rating:0` can be specified in the data object

Good point!

jyash97:
They are getting set on `URL` as : `http://localhost:8001/?RatingsSensor=%5B3%2C5%2Ctrue%5D`

`true` is the value passed to `includeUnrated`.
It is applying on reloading the URL when `URLParams` is `true`

bietkul:
@jyash97 Can we close it now?
I think it has been done here #1003

jyash97:
Yeah this can be closed this was meant for v2. Already fixed in #1003

PR #567, opened 2018-10-09
Fix use of this in static method in Date component

jyash97:
`defaultQuery` was using a function `formatDate` with `this` keyword, which was throwing an error.


packages/web/src/components/date/DatePicker.js
```diff
@@ -23,6 +22,7 @@ import Title from '../../styles/Title';
 import Flex from '../../styles/Flex';
 import CancelSvg from '../shared/CancelSvg';
 import { connect } from '../../utils';
+import formatDate from './utils';
```
metagrover: We should put `formatDate` in core to enable reusability across native components as well.

@divyanshu013 what do you think?

divyanshu013: Agreed bro :handshake:

metagrover: 🤝

jyash97: Need to move to `reactivecore` ?

metagrover: yes, reactivecore utils.

PR #572, opened 2018-10-12
web: Custom routing

axeff:
I've had problems using URLParams with next-router. The history.back() functionality didn't work as expected. So I updated the library to be able to to oen routing via callback function that accepts URLSearchParams as param.


metagrover:
Hi @axeff, thanks for submitting the PR. This looks good, but if you can share a simple replication to your use-case (an example app repo would be ideal), that'd be great.

We just want to make sure that this is the ideal way to fix the issues you're seeing.

PR #586, opened 2018-10-26
vue: Reactive Search For VueJS

bietkul:


packages/vue/.eslintrc
```diff
@@ -0,0 +1,21 @@
+{
+    "env": {
```
metagrover: why aren't we using the linting rules from the root?

bietkul: because different plugins are needed for vue

PR #587, opened 2018-10-26
vue: Move vue-types from reactivecore to vue

bietkul:


metagrover:
@bietkul can you fix these conflicts?

PR #589, opened 2018-10-29
Fix: pass showFilter value based on props

### LINKED ISSUE #588: RatingFilters are unavailable on SelectedFields
Hi there,
I would like to have the RatingsFilter show on the SelectedFields component, but I found it to be impossible because of the following code:
https://github.com/appbaseio/reactivesearch/blob/7625239804b55ede0efa90a2977f1e81072ae3f6/packages/web/src/components/range/RatingsFilter.js#L136-L143

The `showFilter` is always false independently of what's passed in the props
Is it supposed to be like this or is it a bug?

Thanks

jyash97:
Fix #588

### DIFF
--- a/packages/web/src/components/range/RatingsFilter.js
+++ b/packages/web/src/components/range/RatingsFilter.js
@@ -138,7 +138,7 @@ class RatingsFilter extends Component {
 			query: query(value, props),
 			value,
 			label: props.filterLabel,
-			showFilter: false,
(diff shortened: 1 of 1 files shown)


jyash97:
Yeah I tried with web/example/RatingFilters app, and everything seems to work ReactiveList also updates the data correctly. Here is the demo
ratingsfilter

Issue #591, opened 2018-10-30
web: DataSearch and CategorySearch is missing renderSuggestions on the interface


hbish:
**Issue Type:**
bug

**Platform:**
Web

**Description:**
As per the documentation on [Custom Suggestions](https://opensource.appbase.io/reactive-manual/advanced/customsuggestions.html#custom-suggestions-for-datasearch), `renderSuggestion` is exposed as a property on both `DataSearch` and `CategorySearch`.

This is currently not exposed through the interface

**Minimal reproduction of the problem with instructions:**
When trying to define `renderSuggestions` in `DataSearch` component in a react typescript app, the following error is observed.

```
/Users/benshi/prj/my-app/src/components/search/SearchHeader.tsx
(94,17): Property 'renderSuggestions' does not exist on type 'IntrinsicAttributes & DataSearchProps & { children?: ReactNode; }'.
```

**Reactivesearch version:** 
2.12.1 

**Workaround**
Currently it is possible to workaround by adding `renderSuggestions: (...args: any[]) => any;` to the d.ts file

PR #592, opened 2018-10-30
add renderSuggestions as props on DataSearch and CategorySearch (#591)

hbish:
This PR is to resolve missing props on `DataSearch` and `CategorySearch` as documented in #591


metagrover:
Thanks for contributing @hbish 💯  

We will roll this out in the next release 🎉

PR #603, opened 2018-11-05
Add page number and length of displayed results in onResultStats

lakhansamani:
* Returns current page and displayed result stats in `onResultStats`


packages/web/src/components/result/ReactiveList.js
```diff
@@ -411,7 +411,21 @@ class ReactiveList extends Component {
...
+			const streamResults = parseHits(this.props.streamHits) || [];
+			let filteredResults = results;
+
+			if (streamResults.length) {
+				const ids = streamResults.map(item => item._id);
+				filteredResults = filteredResults.filter(item => !ids.includes(item._id));
+			}
+
+			return this.props.onResultStats(
+				this.props.total,
```
metagrover: It would be best if we return an object instead:

```js
const stats = {
    totalPages: <>,
    currentPage: <>,
    ...
}

this.props.onResultStats(stats);
```

This will improved the DX significantly.

lakhansamani: True, lets do that 👍

packages/web/src/components/result/ReactiveList.js
```diff
@@ -420,12 +420,14 @@ class ReactiveList extends Component {
...
 			}
 
-			return this.props.onResultStats(
-				this.props.total,
-				this.props.time,
-				this.state.currentPage,
-				[...streamResults, ...filteredResults].length,
-			);
+			const stats = {
+				totalResults: this.props.total,
```
metagrover: We should also return `totalPages` in addition to this.

lakhansamani: True that will complete all `stats` making that change 👍

PR #617, opened 2018-11-14
web: Add parseDate prop in Date Components

jyash97:
Usage:

```js
<DatePicker
    componentId="DateSensor"
    dataField="date_from"
    parseDate={(value,props) => value.toString('yyyyMMdd')}
/>
```


metagrover:
@jyash97 Can you please resolve conflicts here?

PR #626, opened 2018-11-23
Upgrade lifecycle method & Add renderData, renderAllData

jyash97:
@metagrover Is this correct way to use `renderData` ? How we are planning to use `onData` , I would add its support in this.

### DIFF
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -118,48 +118,48 @@ class ReactiveList extends Component {
 		}
 	}
 
-	componentWillReceiveProps(nextProps) {
-		const totalPages = Math.ceil(nextProps.total / nextProps.size) || 0;
+	componentDidUpdate(prevProps) {
+		const totalPages = Math.ceil(this.props.total / this.props.size) || 0;
 
 		if (
-			!isEqual(this.props.sortOptions, nextProps.sortOptions)
-			|| this.props.sortBy !== nextProps.sortBy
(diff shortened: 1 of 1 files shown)


packages/web/src/components/result/ReactiveList.js
```diff
@@ -118,48 +118,48 @@ class ReactiveList extends Component {
...
 				options.sort = [{
-					[nextProps.dataField]: {
-						order: nextProps.sortBy,
+					[this.props.dataField]: {
+						order: this.props.sortBy,
 					},
 				}];
 			}
-			this.props.setQueryOptions(this.props.componentId, options, true);
+			prevProps.setQueryOptions(prevProps.componentId, options, true);
```
metagrover: It was correct earlier. This is buggy.

packages/web/src/components/result/ReactiveList.js
```diff
@@ -176,40 +176,40 @@ class ReactiveList extends Component {
 			});
 		}
 
-		if (this.props.stream !== nextProps.stream) {
-			this.props.setStreaming(nextProps.componentId, nextProps.stream);
+		if (prevProps.stream !== this.props.stream) {
+			prevProps.setStreaming(this.props.componentId, this.props.stream);
```
metagrover: I feel you have simply replaced `this.props` -> `prevProps` and `nextProps` -> `this.props`. This won't work. Please test all the edge cases properly.

metagrover:
Closing this PR due to incorrectness.

PR #635, opened 2018-11-29
vue: Add ResultCard

jyash97:
Do check the problem that why `ResultCard` was not working on default Export. I tested this using named export and manually Registered this component in Storybook.


packages/vue/src/components/result/ResultCard.jsx
```diff
@@ -0,0 +1,102 @@
...
+		componentId: types.stringRequired,
+		dataField: types.stringRequired,
+		defaultQuery: types.func,
+		excludeFields: types.excludeFields.def([]),
+		innerClass: types.style,
+		listClass: VueTypes.string.def(''),
+		loader: types.title,
+		onAllData: types.func,
+		onData: types.func,
+		onResultStats: types.func,
```
bietkul: Do the same changes here as well as discussed in ResultList PR.

PR #636, opened 2018-11-29
vue: Add ResultList Component

jyash97:
`Note`: I am not able to use default export in Storybook, i have to manually register component on storybook to test them.


packages/vue/src/components/result/ResultList.jsx
```diff
@@ -0,0 +1,113 @@
...
+						...props,
+						listClass: container,
+					}
+				}}
+			/>
+		);
+	},
+	methods: {
+		renderAsList({ item, triggerClickAnalytics }) {
+			const result = this.$props.onData(item);
```
bietkul: @jyash97 Also use scopedSlots here.

jyash97: Like a condition if `scopedSlot` is not defined we can use `$props` ?

bietkul: Yes, similar to reactive list.
```
const onData = this.$scopedSlots.onData || this.$props.onData;
```

jyash97: Got it, but I am not aware, how scopedslot would return object?

packages/vue/src/components/result/ResultList.jsx
```diff
@@ -0,0 +1,113 @@
...
+		stream: types.bool,
+		URLParams: VueTypes.bool.def(false),
+		target: VueTypes.string.def('_blank')
+	},
+	render() {
+		const { onData, ...props } = this.$props;
+		return (
+			<RLConnected
+				{...{
+					props: {
```
bietkul: What's the need to write it like this, is simple props passing not working here?
```
<RLConnected
   listClass={container}
  {...on: {  data: this.renderAsList }}
/>
```
should work

packages/vue/src/components/result/ResultList.jsx
```diff
@@ -0,0 +1,113 @@
...
+		componentId: types.stringRequired,
+		dataField: types.stringRequired,
+		defaultQuery: types.func,
+		excludeFields: types.excludeFields.def([]),
+		innerClass: types.style,
+		listClass: VueTypes.string.def(''),
+		loader: types.title,
+		onAllData: types.func,
+		onData: types.func,
+		onResultStats: types.func,
```
bietkul: We also support scopedSlot for it ( `onResultStats` ) in ReactiveList, so remember to pass it down to ReactiveList.

packages/vue/src/components/result/ResultList.jsx
jyash97: I tried this it wasn't working , will confirm once otherwise we can keep it like this as stated in the plug-in docs

bietkul: ok

jyash97: I tried this it wasn't working , will confirm once otherwise we can keep it like this as stated in the plug-in docs

jyash97: We also need to pass other props using destructuring so we would need props eventually for that as well

packages/vue/src/components/result/ResultList.jsx
```diff
@@ -0,0 +1,113 @@
...
+				);
+			}
+
+			return null;
+		}
+	}
+}
+
+ResultList.install = function (Vue) {
+	Vue.component(ResultList.name);
```
bietkul: I got the issue why you're not able to export components because you're not installing the component.
This line should be
```js
Vue.component(ResultList.name, ResultList);
```
Make that change in ResultCard also.

jyash97: Oh yes, thanks I think I missed it, this should fix things 👍

PR #642, opened 2018-11-29
Vue: rangeSlider Added

divyanshu-rawat:
Basic Layout for Vue:RangeSlider Added.

rangeslider


packages/vue/src/components/range/RangeSlider.jsx
```diff
@@ -0,0 +1,200 @@
...
+    title: types.title,
+    URLParams: VueTypes.bool.def(false)
+  },
+
+  methods:{
+
+    setReact(props) {
+      if (props.react) {
+        this.watchComponent(props.componentId, props.react);
+      }
```
jyash97: In the else part `this.watchComponent(props.componentId, {})` ?

divyanshu-rawat: sure, but any specific reason for doing so?

jyash97: I needed it in dynamicslider, to watch for componentId if react prop was not defined.
@bietkul Can give more details on this

PR #644, opened 2018-11-30
Bug fixes

divyanshu-rawat:
Bug fixes to make defaultSelected story to work.


packages/vue/src/components/range/RangeSlider.jsx
```diff
@@ -129,14 +130,22 @@ const RangeSlider = {
...
+
     this.addComponent(this.$props.componentId);
+    this.addComponent(this.internalComponent);
     this.setReact(this.$props);
-    if (this.selectedValue) {
-      this.handleChange(this.selectedValue);
-    } else if (this.$props.defaultSelected) {
-      this.handleChange(this.$props.defaultSelected);
+
+    const { selectedValue, defaultSelected } = this.$props;
```
bietkul: How you're accessing this `selectedValue` from `this.$props`, afaik it's coming from redux.

divyanshu-rawat: @bietkul yes it is coming as a prop from here - 
```
  selectedValue: state.selectedValues[props.componentId]
    ? state.selectedValues[props.componentId].value
    : null,
```

divyanshu-rawat:
story

packages/vue/src/components/range/RangeSlider.jsx
divyanshu-rawat: Updated

packages/vue/src/components/range/RangeSlider.jsx
```diff
@@ -131,14 +130,23 @@ const RangeSlider = {
     this.setQueryListener(this.$props.componentId, onQueryChange, null);
   },
   beforeMount(){
+
     this.addComponent(this.$props.componentId);
+    this.addComponent(this.internalComponent);
```
jyash97: What is `this.internalComponent` doing? It is not initialized

divyanshu-rawat: Sure,then you can git rid of it.

Issue #584, opened 2018-10-25
Loading information for ReactiveComponent?


csepulv:
I have multiple custom components that use `ReactiveComponent`. Is there a way to get `isLoading` information for these components? I tried to `connect` to the redux store, but I don't think anything but a `ReactiveList` gets `isLoading` info. 

I hope I am doing something wrong as I'd like a way to provide user feedback while the component is waiting on a query.

Thanks

metagrover:
Thanks @csepulv! This is a known issue and will be included in `v3.0.0`

Related issues: https://github.com/appbaseio/reactivesearch/issues/227 https://github.com/appbaseio/reactivesearch/issues/250 https://github.com/appbaseio/reactivecore/issues/12

csepulv:
Thanks. Sorry to have missed it. I can close this since the other issues track this, or leave it open till 3.0. (Let me know your preference)

bietkul:
Check `isLoading` property in child component's props, it can be used to render a loader component.
https://opensource.appbase.io/reactive-manual/v3/advanced/reactivecomponent.html#child-component

Issue #250, opened 2018-02-07
How can I add element in initial Loader props??


ashwini987:
**Issue Type:**
[Enhancement]

**Description:**
In Version 1, I can use initial-Loader props which value I provide as a string.
Now I want to add spinner or Loader instead of string value.
How can I add extra element in initial-Loader.
And in Version 2 no such a props that work as a initial-loader.

**Screenshots:**
![download](https://user-images.githubusercontent.com/31088567/35899988-88c8cc6a-0bf5-11e8-8031-da8746144315.png)

**Reactivesearch version:v1/v2

**Browser:** [all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

divyanshu013:
Hi @ashwini987, in v2 right now the `loader` support is only available in the result components. We'll be adding the support in other components soon. Related issue https://github.com/appbaseio/reactivesearch/issues/227

Issue #227, opened 2018-01-23
Add loader support


divyanshu013:
**Issue Type:**

Enhancement (Both web and native)

**Description:**

We should add support for `loader` prop in components which fetch data asynchronously. This sometimes causes the components to appear a bit after the entire page has loaded and push the content which looks bad IMO.

**Minimal reproduction of the problem with instructions:**

Try any app built with reactivesearch (say [gitxplore-app](https://appbaseio-apps.github.io/gitxplore-app/)) and notice that the filters pop in after the UI has loaded. Throttling the network should also help in replicating the issue.

**Reactivesearch version:** x.y.z

2.2.0

**Browser:** [all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

All

siddharthlatest:
- [ ] Add loader support for ReactiveComponent. (Request reference: https://gitter.im/appbaseio/reactivesearch?at=5b5b8c8f17c942036b79dfc8)

PR #649, opened 2018-12-02
web: Add loader support

bietkul:
This PR has the following fixes.
https://github.com/appbaseio/reactivesearch/issues/227
https://github.com/appbaseio/reactivesearch/issues/250
https://github.com/appbaseio/reactivesearch/issues/584

Features
- Add loader in search components
- Add loader in range components
- Add loader in list & base components
- Provide `isLoading` as a prop in `ReactiveComponent` child

### DIFF
--- a/packages/web/src/components/basic/ReactiveComponent.js
+++ b/packages/web/src/components/basic/ReactiveComponent.js
@@ -130,7 +130,8 @@ class ReactiveComponent extends Component {
 
 		try {
 			const childrenWithProps = React.Children.map(children, child =>
-				React.cloneElement(child, { ...rest, setQuery: this.setQuery }));
+				React.cloneElement(child, { ...rest, setQuery: this.setQuery }),
+			);
 			return <div>{childrenWithProps}</div>;
 		} catch (e) {
 			return null;
@@ -152,6 +153,7 @@ ReactiveComponent.propTypes = {
 	watchComponent: types.funcRequired,
 	aggregations: types.selectedValues,
 	hits: types.data,
+	loading: types.bool,
 	selectedValue: types.selectedValue,
 	// component props
 	children: types.children,
@@ -173,6 +175,7 @@ const mapStateToProps = (state, props) => ({
 		(state.selectedValues[props.componentId]
 			&& state.selectedValues[props.componentId].value)
 		|| null,
+	loading: state.isLoading[props.componentId],
 });
 
 const mapDispatchtoProps = dispatch => ({


packages/web/src/components/basic/ReactiveComponent.js
```diff
@@ -173,6 +175,7 @@ const mapStateToProps = (state, props) => ({
 		(state.selectedValues[props.componentId]
 			&& state.selectedValues[props.componentId].value)
 		|| null,
+	loading: state.isLoading[props.componentId],
```
metagrover: Should we rename it to `isLoading` everywhere?

packages/web/src/components/list/MultiDropdownList.d.ts
```diff
@@ -23,6 +23,7 @@ export interface MultiDropdownList extends CommonProps {
 	size?: number;
 	sortBy?: types.sortByWithCount;
 	themePreset?: types.themePreset;
+	loader: any;
```
metagrover: I think `loader` can be of the same type as that of `title`.

packages/web/src/components/list/SingleDropdownList.js
```diff
@@ -249,6 +249,10 @@ class SingleDropdownList extends Component {
 		const { isLastBucket } = this.state;
 		let selectAll = [];
 
+		if (this.props.loading && this.props.loader) {
```
metagrover: Would it better if we create a helper util for this - say `renderLoader` which takes the `props` and renders the loader when needed?

packages/web/src/components/search/CategorySearch.js
```diff
@@ -489,6 +489,25 @@ class CategorySearch extends Component {
 		return null;
 	}
 
+	renderLoader = () => {
+		const {
+			loader, loading, themePreset, theme,
+		} = this.props;
+		const { currentValue } = this.state;
+		if (!(loading && loader && currentValue)) {
```
metagrover: Ideally, we should write:

```
if (loading && loader && currentValue) {
    return (...);
}

return null;
```

to improve code readability here.

packages/web/src/components/search/DataSearch.js
```diff
@@ -446,6 +446,25 @@ class DataSearch extends Component {
 		return null;
 	}
 
+	renderLoader = () => {
+		const {
+			loader, loading, themePreset, theme,
+		} = this.props;
+		const { currentValue } = this.state;
+		if (!(loading && loader && currentValue)) {
```
metagrover: Same as before.

metagrover:
This will be available in v2 and v3.

packages/web/src/components/basic/ReactiveComponent.js
bietkul: It won't effect anyways, since we're not exposing it to consumers.

metagrover: It improves the readabilty and meaning.

packages/web/src/components/list/SingleDropdownList.js
bietkul: but still we have to write the condition to render the loader.
Can you please describe here how we can use a util function here & what'll be the advantages?

metagrover: Sure, we can import that util method in every component:

```js
import { renderLoader } from '...';

// inside component render:
render() {
    return (
        <div>
            {renderLoader(props)}
        </div>
    )
}
```

Now, we don't have to repeat the check for `props.isLoading && props.loader` everywhere - since the rendering of the loader is completely dependent on `props`.

bietkul: But that's not the case, we just need to render `loader` ( nothing else so we have to write a check before renderLoader ) if `isLoading` is `true`.

metagrover: Not sure if I am super clear on that. Here is what I have in mind:

```js
const renderLoader = (props) => {
    if (props.isLoading && props.loader) {
        return props.loader;
    }
    return null;
}
```

So we are checking `isLoading` & `loader` props here in every component right?

bietkul: that's ok, just want to do the same but the problem is that we still need to write the check to determine if `isLoading` true then only we have to render the loader.
So, the code will be like that with that snippet
```js
render() {
   if (isLoading) { return renderLoader(this.props); };
   // other code
}
```
which doesn't has any advantage

metagrover: um, we are checking for `isLoading` inside the `renderLoader` util, so why check it before calling the function?

PR #651, opened 2018-12-03
vue: Add labels in RangeSlider

jyash97:
image

### DIFF
--- a/packages/vue/src/components/range/RangeSlider.jsx
+++ b/packages/vue/src/components/range/RangeSlider.jsx
@@ -32,10 +32,7 @@ const RangeSlider = {
 			start: VueTypes.integer.def(0),
 			end: VueTypes.integer.def(10),
 		}),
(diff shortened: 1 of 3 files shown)


packages/vue/src/components/range/RangeSlider.jsx
```diff
@@ -32,6 +32,10 @@ const RangeSlider = {
 			start: VueTypes.integer.def(0),
 			end: VueTypes.integer.def(10),
 		}),
+		rangeLabels: VueTypes.shape({
```
bietkul: Do we not have this prop type in `types` list?

divyanshu-rawat: @bietkul I guess `rangeLabels`  prop is only for `rangeSliders` we don't have this in `types` list, we have `missingLabel` and `selectAllLabel` but they are used when we select any `filter`.

PR #658, opened 2018-12-05
Add showCount and renderListItem to Single/MultiDataList

### LINKED ISSUE #270: Add showCount property to SingleDataList and MultiDataList
**Issue Type:**
enhancement

**Description:**
Just like SingleList and MultiList, SingleDataList and MultiDataList could make use of a showCount property. They can update asynchronously so the counts will only appear when loaded without disrupting the "static" nature of of the DataList labels. There can be a separate option on whether labels that return a 0 count should be hidden or not.

This can be especially useful as an alternative to SingleList and MultiList because sometimes the aggregation counts are quite slow, when compared to the actual search query. By using a hybrid approach of static label filters with dynamic counts, users don't have to wait for the counts to load before receiving search results or filter options.

lakhansamani:
- Update examples to use new meetup_app

Resolves #270

### DIFF
--- a/packages/web/src/components/list/MultiDataList.js
+++ b/packages/web/src/components/list/MultiDataList.js
@@ -260,20 +260,24 @@ class MultiDataList extends Component {
 
 	updateStateOptions = (bucket) => {
 		if (bucket) {
+			const bucketDictionary = bucket.reduce(
+				(obj, item) => ({
+					...obj,
+					[item.key]: item.doc_count,
+				}),
+				{},
+			);
+
 			const { options } = this.state;
-			const newOptions = [];
-
-			options.forEach((item) => {
-				const doc = bucket.find(bucketItem => bucketItem.key === item.value);
-
-				if (doc) {
-					newOptions.push({
+			const newOptions = options.map((item) => {
+				if (bucketDictionary[item.value]) {
+					return {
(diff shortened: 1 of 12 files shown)


packages/web/src/components/list/MultiDataList.js
```diff
@@ -214,6 +247,41 @@ class MultiDataList extends Component {
...
+		props.setQueryOptions(this.internalComponent, queryOptions);
+	};
+
+	updateStateOptions = (bucket) => {
+		if (bucket) {
+			const { options } = this.state;
+			const newOptions = [];
+
+			options.forEach((item) => {
+				const doc = bucket.find(bucketItem => bucketItem.key === item.value);
```
metagrover: This looks performance intensive. Can we improve this a little bit? `find` will run on every iteration. Maybe we can flatten the buckets into a dictionary to bring the traversal complexity to O(1).

metagrover: Also, I'd recommend you to use `map` here instead of `forEach` and `push`, it will improve the readability further 😸

metagrover:
Since, SSR doesn't add much value here, we won't be supporting SSR (showing count) for these components until a valid use-case/request comes up.

PR #672, opened 2018-12-07
web: Remove DOM element if data is not present in ResultCard

jyash97:


metagrover:
@jyash97 Same for `next` please 🐱

metagrover:
@jyash97 This will be applicable to `ResultList` as well.

jyash97:
Done 🏁

PR #690, opened 2018-12-14
web: Richer support for queryDSL in defaultQuery and customQuery props

### LINKED ISSUE #546: Enhancement: Richer support for queryDSL in defaultQuery and customQuery props
**Issue Type:**

Enhancement

**Platform:**

ReactiveSearch for web, native, maps

**Description:**

Currently, there are two props: `defaultQuery` and `customQuery` that can be used for customizing a component's data-driven query. The issue requests consistency between the two props and accepting all the fields a user passes to be treated as top-level fields.

E.g. currently, a user can't pass something like [`track_scores`](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/search-request-sort.html#_track_scores) as we only accept `query`, `aggs` and `sort` as the top level fields in defaultQuery. customQuery doesn't even read `aggs` or `sort` field values.

What this issue proposes is to have ReactiveSearch validate the fields it uses internally (`query`, `aggs`, `sort`), but accept other fields that a user provides inside the defaultQuery or customQuery prop. If however, a user only provides a direct query, ReactiveSearch can treat as the value of the `query` key.

Here are two examples that should work with both these props as per this issue.

```
customQuery={{
  track_scores: true,
  query: {
    match: {
      my_field: "foo"
    }
  }
}}
```
In the above, we take `track_scores` as a top-level field as it's present along with the `query` field.

```
defaultQuery={{
  term: {
    my_field: "foo"
  }
}}
```
In the above, we treat the entire `term` query to be a value of the `query` key as it's not present.

**Screenshots:**

2.10.1

All

**Anything else:**

bietkul:
Fixes #546

### DIFF
--- a/packages/web/src/components/basic/DataController.js
+++ b/packages/web/src/components/basic/DataController.js
@@ -5,8 +5,13 @@ import {
 	removeComponent,
 	updateQuery,
 	setQueryListener,
+	setQueryOptions,
 } from '@appbaseio/reactivecore/lib/actions';
-import { isEqual, checkValueChange } from '@appbaseio/reactivecore/lib/utils/helper';
+import {
+	isEqual,
+	checkValueChange,
+	getOptionsFromQuery,
+} from '@appbaseio/reactivecore/lib/utils/helper';
 
 import types from '@appbaseio/reactivecore/lib/utils/types';
 
@@ -46,7 +51,17 @@ class DataController extends Component {
 
 	updateQuery = (defaultSelected = null, props) => {
 		this.locked = true;
-		const query = props.customQuery || DataController.defaultQuery;
+		const { customQuery } = props;
+		const query = customQuery || DataController.defaultQuery;
+
+		const customQueryOptions = customQuery
+			? getOptionsFromQuery(customQuery(defaultSelected, props))
+			: null;
+		this.queryOptions = {
+			...this.queryOptions,
+			...customQueryOptions,
+		};
+		props.setQueryOptions(props.componentId, this.queryOptions);
 
 		const performUpdate = () => {
 			props.updateQuery({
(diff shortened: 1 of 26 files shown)


packages/web/src/components/basic/DataController.js
```diff
@@ -46,7 +51,17 @@ class DataController extends Component {
...
-		const query = props.customQuery || DataController.defaultQuery;
+		const { customQuery } = props;
+		const query = customQuery || DataController.defaultQuery;
+
+		const customQueryOptions = customQuery
+			? getOptionsFromQuery(customQuery(defaultSelected, props))
+			: null;
+		this.queryOptions = {
+			...this.queryOptions,
+			...customQueryOptions,
```
metagrover: This won't work if `customQueryOptions` is `null` as per Line 59.

bietkul: @metagrover Can you please describe what'll be the problem if `customQueryOptions` is `null`, because if it's `null` then still the `queryOptions` will be an empty object.

metagrover: If customQuery is null, we can't spread it - `{ ...null }` won't work.

metagrover: My bad. This works!

metagrover:
@bietkul We should support this behavior with `defaultQuery` prop as well. Some components like ReactiveList, ResultList, ResultCard and ReactiveComponent (and maybe search components as well) support `defaultQuery` prop atm.

bietkul:
@metagrover I'm already handling it.

metagrover:
I think you missed out on `ReactiveComponent`.

bietkul:
> I think you missed out on `ReactiveComponent`.

Fixed it.

Issue #595, opened 2018-10-31
Feature: Add support for `nested` object mapping


siddharthlatest:
**Issue Type:**

Enhancement

**Platform:**

ReactiveSearch for Web

**Description:**

ReactiveSearch today doesn't support querying the fields of an object field with a [nested mapping](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/nested.html).

A nested object requires a different query (where a path for the nested field needs to be specified) along with the original query. The same applies to an aggregation.

This issue proposes to add a `nestedField` prop in all the data-driven components we have to support the use-case of a nested object.

Prop definition for the field looks like:

**nestedField** `String` [optional]  
    When your dataField is present inside an object of type [`nested`](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/nested.html), this can be used to set the path of the nested field. Setting this allows searching within documents belonging to the path. 

**Reactivesearch version:** x.y.z

2.12.1

metagrover:
@aol-nnov Left the reviews on the commit. Thanks for taking out the time to work on this!

aol-nnov:
Thanks, I'll revisit the changes soon.

It worked for me months before, but I was never confident if it's a clear fix for nested fields.

aol-nnov:
@metagrover I've redone my changes in `MultiList`.

My previous idea was to store "nested" part of property name in `nestedField` 
i.e. `vendor` is nested object with field `name`
```
dataField: 'name'
nestedField: 'vendor'
```
(it was some time before your changes for nested mapping support)

Now, I've followed your guidelines and made it your way:
```
dataField: 'vendor.name'
nestedField: 'vendor'
```

tested it with my local ES instance of my project. Everyting works like a charm! Finally I do not need a patched reactivesearch, hooray! ;)

metagrover:
Probably something's off with the `defaultQuery`. Check if the final query is being wrapped by `nested` query?

I'm looking into this!

aol-nnov:
no need to.. It was me again ). I've updated the comment (strike through was not applied though)

TCMiranda:
Hey guys, thanks for this, really helpful!
Here follows an observation: One `nested` property is allowed to have another `nested` one inside it. This API doesn't support that case.

Since it is a new API, wouldn't it be appropriate to consider an API definition which enables this "recursion" into the appropriate nested level?

I believe its a corner case, but i've come to it recently:
```
{
  nested: {
    path: 'trajectory',
[...]

PR #694, opened 2018-12-14
Add nestedField Support in base Components

jyash97:
#595


packages/web/src/components/basic/TagCloud.js
```diff
@@ -204,15 +216,35 @@ class TagCloud extends Component {
 	static generateQueryOptions(props) {
 		const queryOptions = getQueryOptions(props);
 		queryOptions.size = 0;
-		queryOptions.aggs = {
-			[props.dataField]: {
-				terms: {
-					field: props.dataField,
-					size: props.size,
-					order: getAggsOrder(props.sortBy || 'asc'),
+		const { nestedField } = props;
```
metagrover: We should use `getAggsQuery(queryOptions, props)` util here. Check `SingleList` component for further implementation details.

packages/web/src/components/basic/utils.js
```diff
@@ -0,0 +1,39 @@
+import { getAggsOrder } from '@appbaseio/reactivecore/lib/utils/helper';
+
+const getAggsQuery = (query, props) => {
```
metagrover: This utility already exists. Why have we recreated it here?

jyash97: There were some query related changes in the `list components` like - `includesQuery`, `missing`, so I thought to keep both the utils specific to the type of components ( `base` &  `list` )

metagrover: `TagCloud` is also a list component underneath.

jyash97: Okay will try & update this

Issue #706, opened 2018-12-18
VueRs: MultiDropdownList emitting single value.


divyanshu-rawat:
**Issue Type:**
BUG

**Platform:**
version - 1.0.0-beta.1

**Description:**
Vue ReasctiveSearch MultiDropdownList emitting single selected value, instead of returning an array of selected values.

jyash97:
@divyanshu-rawat Is this issue solved in #708  ?

PR #707, opened 2018-12-18
VueRs:fixed MultiDropdownList single value emit.

divyanshu-rawat:
VueRs: MultiDropdownList now emits an `array` of selected values, instead of returning a `singleselectedvalue`.


divyanshu-rawat:
#706 fix @bietkul.

PR #708, opened 2018-12-18
Multi range example

divyanshu-rawat:
1 - Added example for `multiRange`.
2 - VueRs: MultiDropdownList now emits an `array of selected values`, instead of returning a single `selectedvalue`.


divyanshu-rawat:
fixes #706  @bietkul 👍

PR #712, opened 2018-12-19
web: Add nestedField Support in Search Components

jyash97:
Category Search Demo
categorynested

DataSearch Demo
datasearchnested


packages/web/src/components/search/CategorySearch.js
```diff
@@ -350,6 +350,17 @@ class CategorySearch extends Component {
 				},
 			} // prettier-ignore
 			: query(value, props, category);
+
+		if (queryObject && props.nestedField) {
```
metagrover, line 354: We should do this nested query wrapping inside the `static defaultQuery` method of the search components.

---

FYI `props.defaultQuery` is used for a very particular use-case in data-driven components. The current implementation is faulty and will be updated in our future release.

jyash97:
Fixed in #759

PR #713, opened 2018-12-20
Remove input validation onChange of slider

lakhansamani:
**Issue**
- While typing value in end input, it doesn't allow the user to enter value less than start value, because `RangeSlider` change handler resets it to start value.

ejtxgox16y

### DIFF
--- a/packages/web/src/components/range/RangeInput.js
+++ b/packages/web/src/components/range/RangeInput.js
@@ -22,8 +22,8 @@ class RangeInput extends Component {
 			isEndValid: true,
 		};
 
-		this.startInputRef = null;
-		this.endInputRef = null;
+		this.startInputRef = React.createRef();
+		this.endInputRef = React.createRef();
 	}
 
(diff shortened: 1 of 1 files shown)


metagrover:
I'm not sure if I understand this completely. Is this the issue you are seeing or the fix you have implemented?

---

The rangeslider component inputs should not allow setting values out of bounds.

packages/web/src/components/range/RangeInput.js
```diff
@@ -21,6 +21,9 @@ class RangeInput extends Component {
 			isStartValid: true,
 			isEndValid: true,
 		};
+
+		this.startInputRef = null;
```
metagrover: Please use `React.createRef()` API to create refs.

metagrover:
@lakhansamani We will need this on `next` as well 🐱

lakhansamani:
> @lakhansamani We will need this on `next` as well 🐱

Okay creating PR for it

PR #725, opened 2018-12-21
Fix range input validation onChange of slider

lakhansamani:


packages/web/src/components/range/RangeInput.js
```diff
@@ -21,11 +21,14 @@ class RangeInput extends Component {
...
 		};
+
+		this.startInputRef = React.createRef();
+		this.endInputRef = React.createRef();
 	}
 
 	componentDidUpdate(prevProps) {
-		if (!isEqual(this.props.value, prevProps.value)) {
-			this.handleSlider(this.props.value);
+		if (!isEqual(this.props.defaultValue, prevProps.defaultValue)) {
```
metagrover: We should only watch for `value` prop updates. `defaultValue` should only be used on initial mount.

PR #734, opened 2018-12-27
vue: Add support for Dynamic Range Slider

### LINKED ISSUE #722: Feature: Aggregate min and max in RangeSlider
Not sure if I am missing something or if it is developed in the react libraries and not in Vue but it would be awesome if the `RangeSlider` could aggregate the min and max values of the field that was passed automatically.

This would mean that the min and max values are not needed anymore.

There are no docs for that component so I am not 100% sure but I couldn't find anything related to that [in the code](https://github.com/appbaseio/reactivesearch/blob/dev/packages/vue/src/components/range/RangeSlider.jsx).

jyash97:
Usage: 
```js
     <DynamicRangeSlider
          dataField="books_count"
          componentId="BookSensor"
          title="Hey"
          :rangeLabels="function(min, max){
              return {
                start: min + ' Books',
                end: max + ' Books'
              }
            }"
            :defaultSelected="function(min,max){
              return {
                start: min + 100,
                end: max - 100
              }
            }"
        />
```

Preview:
drs

Closes #722


packages/vue/src/components/range/DynamicRangeSlider.jsx
```diff
@@ -0,0 +1,307 @@
...
+import { connect } from '../../utils/index';
+import Title from '../../styles/Title';
+import Slider from '../../styles/Slider';
+import types from '../../utils/vueTypes';
+
+const { addComponent, removeComponent, watchComponent, updateQuery, setQueryListener,setQueryOptions } = Actions;
+
+const { checkValueChange, getClassName } = helper;
+
+const getComponents = () => {
```
bietkul: We can move this function to utils.

jyash97: Sure thing, updating

PR #735, opened 2018-12-27
vue: RangeSlider code cleanup

jyash97:
### What this PR includes
If `range` is not defined, the current implementation breaks the component. So this PR solves this, instead of rendering `Null` we can also render error message like `range prop is not defined`. I am also putting a console error when `range` is not defined.

In this way, the user will see the title if defined but won't see the slider.

### Preview:
Check console message.
image

cc @divyanshu-rawat @bietkul


bietkul:
@jyash97 It'll be better to implement `onError` & `renderError` otherwise `console.error` is enough.
No need to render any hardcoded error message.

jyash97:
Great updating the PR

PR #740, opened 2018-12-28
web: Fix controlled behaviour of Components & other fixes

jyash97:
### What this PR includes -

- Fix control behavior
   - Check for `value === undefined` as value can be empty string which doesn't trigger `onChange`.
   - If `onChange` is not defined, switch to uncontrolled behavior.

- Fix static member access
   - When testing the implementation I noticed `defaultQuery`  is accessed using `this` & hence breaking the component

- RangeSlider Fix
  - When testing RangeSlider, updateQuery wasn't firing on changing the component, have fixed this as well.


packages/web/src/components/range/RangeSlider.js
```diff
@@ -231,23 +231,27 @@ class RangeSlider extends Component {
...
-				} else {
-					// since value prop is set & onChange is not defined
-					// we need to reset the slider position
-					// to the original 'value' prop
-					this.setState({
-						currentValue: this.state.currentValue,
-					});
-				}
+
+			if (value === undefined) {
```
metagrover, line 236: This is wrong. Please compare the logic flow for the code before and after.

metagrover:
@jyash97 can you please resolve conflicts here?

PR #752, opened 2019-01-02
fix #607

### LINKED ISSUE #607: ReactiveBase tries to parse all query parameters as JSON
**Issue Type:**
bug

**Platform:**
web

**Description:**
When creating a new `<ReactiveBase>` it tries to parse all the query parameters as JSON strings.
If it encounters any parameter that fails to parse as JSON, it will stop and log an error to the console.

```
index.js:1 REACTIVESEARCH - An error occured while parsing the URL state. SyntaxError: Unexpected token M in JSON at position 0
    at JSON.parse (<anonymous>)
    at ReactiveBase.js:147
    at Array.forEach (<anonymous>)
    at ProxyComponent.setStore (ReactiveBase.js:144)
    at new ReactiveBase (ReactiveBase.js:67)
```

https://github.com/appbaseio/reactivesearch/blob/7a9fb9bd39213bf4fde0f5268d4c9350ac4bedaa/packages/web/src/components/basic/ReactiveBase.js#L87

- When rendering the `<ReactiveBase>` inside a Storybook (that uses query paramters) it will always log an error.
- If the application has it's own query parameters (that are not JSON), it will not get any values.

**Anything else:**
Remove the `console.error` message - It's not helpful to print the stack trace when a json parse fails.

If it is necessary for the query parameters you use to be JSON encoded, i'd recommend a check before doing `JSON.parse`. Does the value start with `"` could be a simple check. Maybe fallback to the plain value if it isn't a JSON string? So you could do `?query=searchterm`.

Adding the `try...catch` around each parse could also make sense, so even if a parse fails, it will still find the remaining valid values

The option to disable it would also make sense, since you can disable URLParams or search components.

rap2hpoutre:
See #607. The try/catch was at a wrong level. Now, it _ignores_ URL params that are not JSON encoded.

### DIFF
--- a/packages/web/src/components/basic/ReactiveBase.js
+++ b/packages/web/src/components/basic/ReactiveBase.js
@@ -79,18 +79,19 @@ class ReactiveBase extends Component {
 		const params = new URLSearchParams(queryParams);
 		let selectedValues = {};
 
-		try {
-			Array.from(params.keys()).forEach((key) => {
+
+		Array.from(params.keys()).forEach((key) => {
+			try {
+				const value = JSON.parse(params.get(key));
 				selectedValues = {
 					...selectedValues,
-					[key]: { value: JSON.parse(params.get(key)) },
+					[key]: { value },
 				};
-			});
-		} catch (e) {
-			console.error('REACTIVESEARCH - An error occured while parsing the URL state.', e);
-			selectedValues = {};
-		}
-
+			} catch (e) {
+				// Do not add to selectedValues if JSON parsing fails.
+			}
+		});
+		
 		const { headers = {}, themePreset } = props;
 		const appbaseRef = Appbase(config);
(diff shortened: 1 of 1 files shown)


metagrover:
@rap2hpoutre Thank you for the contribution 💯

We will roll this out shortly 🎉

rap2hpoutre:
@metagrover Thank you! 👍  I was about to create a manual fix in my app, but I think I should wait... Do you have any idea about when will it be released (on 2.x branch)?

metagrover:
I will definitely do the release later today for `2.x`.

metagrover:
@rap2hpoutre `2.16.0` Release is out. Upgrade to get things rolling! 🎉

rap2hpoutre:
Thank you!!

It still fails because URLParamsProvider has the same issue, sorry I did not saw that. See #754.

PR #754, opened 2019-01-02
web: following #752

rap2hpoutre:
Oops. It still fails for #607 (half-fixed via #752) because URLParamsProvider has the same problem.


metagrover:
Thank you for the contribution! Released in `v2.16.1` 🎉

rap2hpoutre:
Thanks it works!

PR #777, opened 2019-01-11
web: Fix controlled behavior in List  when onChange is not defined

jyash97:
NOTE:

1. When `onChange` is not defined in `SingleDropdownList` and we select a value the dropdown closes because `Dropdown` component uses the state to control dropdown.
2. Currently, on clearing `SelectedFilters` the value is getting reset as we haven't implemented that SelectedFitler behavior in the controlled environment.
3. `SingleDataList`  initial value in controlled behavior.

### DIFF
--- a/packages/web/src/components/list/SingleDataList.js
+++ b/packages/web/src/components/list/SingleDataList.js
@@ -32,7 +32,8 @@ class SingleDataList extends Component {
 	constructor(props) {
 		super(props);
 
-		const currentValue = props.selectedValue || props.defaultValue || '';
+		const defaultValue = props.defaultValue || props.value;
+		const currentValue = props.selectedValue || defaultValue;
 		this.state = {
 			currentValue,
 			searchTerm: '',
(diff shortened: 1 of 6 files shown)


metagrover:
> Currently, on clearing SelectedFilters the value is getting reset as we haven't implemented that SelectedFitler behavior in the controlled environment.

I've already implemented this here - 4ffee7c0dfddbe8b5df0d5b4d1a0a2b7a7dacfe2

packages/web/src/components/list/SingleDataList.js
```diff
@@ -32,7 +32,8 @@ class SingleDataList extends Component {
 	constructor(props) {
 		super(props);
 
-		const currentValue = props.selectedValue || props.defaultValue || '';
+		const defaultValue = props.defaultValue || props.value;
+		const currentValue = props.selectedValue || defaultValue;
```
metagrover: It should be:

```suggestion
		const currentValue = props.selectedValue || defaultValue || '';
```

metagrover (changes requested):

metagrover (approved):

PR #780, opened 2019-01-11
FIx controlled behavior for Date Components

jyash97:
Note: 

If onChange is not defined we can still change the calendar date, it is due to `dayPickerInput`; the component is handling the date value. The query is not fired though.

Demo: 
SeletedFilter are not updated on changing dates because onChange is not defined.
datepickeronchange

### DIFF
--- a/packages/web/src/components/date/DatePicker.js
+++ b/packages/web/src/components/date/DatePicker.js
@@ -109,7 +109,19 @@ class DatePicker extends Component {
 
 	clearDayPicker = () => {
 		if (this.state.currentDate !== '') {
-			this.handleDateChange(''); // resets the day picker component
+			const { value, onChange } = this.props;
+
+			if (value === undefined) {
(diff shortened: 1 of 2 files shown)


packages/web/src/components/date/DatePicker.js
```diff
@@ -120,7 +121,9 @@ class DatePicker extends Component {
 		} else if (onChange) {
 			onChange(date || '');
 		} else {
-			this.handleDateChange(date || '');
+			this.setState(state => ({
```
metagrover, line 138: It would be great if you can add comments here to state the intent for this key update. Future you will thank you 😸 

```
// this will trigger a remount on the date component
// since it doesn't respect the controlled behavior setting on its own
```

metagrover (changes requested):

metagrover (approved):

PR #781, opened 2019-01-11
Check for customQuery before destructuring

jyash97:


packages/web/src/components/basic/DataController.js
```diff
@@ -56,7 +56,8 @@ class DataController extends Component {
 		let query = DataController.defaultQuery(defaultSelected, props);
 		let customQueryOptions;
 		if (customQuery) {
-			({ query } = customQuery(defaultSelected, props));
+			const customQueryObject = customQuery(defaultSelected, props);
+			query = customQueryObject && customQueryObject.query;
```
bietkul: Can't we just do
`({ query } = customQuery(defaultSelected, props) || {})`

bietkul (changes requested):

packages/web/src/components/basic/DataController.js
metagrover: Agree, that's what I'm doing in ReactiveComponent.

jyash97: Sure will update this in files

bietkul (approved):

PR #784, opened 2019-01-14
web: Added defaultQuery in List Components

bietkul:


packages/web/src/components/search/CategorySearch.js
```diff
@@ -410,11 +410,15 @@ class CategorySearch extends Component {
 			} // prettier-ignore
 			: query;
 
-		props.setQueryOptions(this.props.componentId, {
+		const finalQueryOptions = componentId === this.internalComponent ? {
```
metagrover: It's best if we undo this - because we won't be supporting `defaultQuery` in search components.

metagrover (changes requested):

packages/web/src/components/search/CategorySearch.js
bietkul: Updated

metagrover (approved):

PR #788, opened 2019-01-14
web: Fix selecteFilter behavior in List controlled

jyash97:


packages/web/src/components/list/SingleDataList.js
```diff
@@ -57,7 +57,7 @@ class SingleDataList extends Component {
 		this.setReact(props);
 
 		if (currentValue) {
-			this.setValue(currentValue, props, hasMounted);
+			this.setValue(currentValue, false, props, hasMounted);
```
metagrover: It should be `true` imo since it's a default value. Why is it `false` here? 🤔

packages/web/src/components/list/SingleList.js
```diff
@@ -60,7 +60,7 @@ class SingleList extends Component {
 		const hasMounted = false;
 
 		if (currentValue) {
-			this.setValue(currentValue, props, hasMounted);
+			this.setValue(currentValue, false, props, hasMounted);
```
metagrover: Same as before. It should be `true`

metagrover (changes requested):

packages/web/src/components/list/SingleDataList.js
jyash97: Yeah actually I did it to not reset the selectedFilter when `value` is defined and `onChange` is not defined. I think the name of the variable should be changed, something like `changeSelectedFilter`

Thoughts?

metagrover: Interesting! I don't think that adds any value. We have checks in place that allows the SelectedFilter component to only update a value if there is a change. So, it would be best if we treat this update as a defaultValue because then it will always ensure that it follows an optimised flow of logic.

jyash97: Yeah makes sense
Updating this.

metagrover (approved):

PR #796, opened 2019-01-16
vue: Add nestedField support in Range Components

jyash97:


bietkul (approved):

bietkul:
@jyash97 Please update the docs.

PR #800, opened 2019-01-18
vue: Fix data variables use

jyash97:


packages/vue/src/components/range/DynamicRangeSlider.jsx
```diff
@@ -16,12 +16,12 @@ const DynamicRangeSlider = {
 	name: 'DynamicRangeSlider',
 	components: getComponents(),
 	data() {
-		this.state = {
+		this._state = {
```
bietkul: @jyash97 Remove this keyword from here.
```
const state = {...}
```
should be fine.

packages/vue/src/components/range/RangeSlider.jsx
```diff
@@ -16,12 +16,12 @@ const RangeSlider = {
 	name: 'RangeSlider',
 	components: getComponents(),
 	data() {
-		this.state = {
+		this._state = {
```
bietkul: Same changes needed here too.

bietkul (changes requested):

packages/vue/src/components/range/DynamicRangeSlider.jsx
jyash97: Okay doing the changes

bietkul (approved):

PR #811, opened 2019-01-31
Make Tileserver customizable from props

mkaraula:
IWLT propose an additional prop for ReactiveOpenStreetMap do change the tile server if needed. If the prop is not set ReactiveOpenStreetMap will default to `http://{s}.tile.osm.org/{z}/{x}/{y}.png`

### DIFF
--- a/packages/maps/src/components/result/ReactiveOpenStreetMap.js
+++ b/packages/maps/src/components/result/ReactiveOpenStreetMap.js
@@ -160,7 +160,7 @@ class ReactiveOpenStreetMap extends Component {
 				}}
 				touchZoom
 			>
-				<OpenStreetLayer url={this.props.tileServer ? this.props.tileServer : 'http://{s}.tile.osm.org/{z}/{x}/{y}.png'} />
+				<OpenStreetLayer url={this.props.tileServer || 'https://{s}.tile.osm.org/{z}/{x}/{y}.png'} />
 				{markers}
 				{this.props.showMarkers && this.props.markers}
 			</OpenStreetMap>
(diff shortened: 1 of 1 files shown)


packages/maps/src/components/result/ReactiveOpenStreetMap.js
```diff
@@ -160,7 +160,7 @@ class ReactiveOpenStreetMap extends Component {
 				}}
 				touchZoom
 			>
-				<OpenStreetLayer url="http://{s}.tile.osm.org/{z}/{x}/{y}.png" />
+				<OpenStreetLayer url={this.props.tileServer ? this.props.tileServer : 'http://{s}.tile.osm.org/{z}/{x}/{y}.png'} />
```
lakhansamani:
```suggestion
				<OpenStreetLayer url={this.props.tileServer || 'https://{s}.tile.osm.org/{z}/{x}/{y}.png'} />
```

metagrover (approved):

metagrover (changes requested):
Looks good 👌 

One quick modification - we will need to add `tileServer` prop to this component's `propTypes` as well.

mkaraula:
Hey @metagrover, where can I find the Prop Types for this component? I was a bit lost while looking for it 😅

metagrover:
Since we use `ReactiveMap` as the base component for `ReactiveOpenStreetMap`, we don't have additional props (hence propTypes) in this component. So, you can add it as:

```js
// import types:
import types from '@appbaseio/reactivecore/lib/utils/types';

// add proptypes:
ReactiveOpenStreetMap.propTypes = {
    tileServer: types.string,
};
```

in this file.

metagrover (approved):

PR #817, opened 2019-02-05
Fix: target prop should be optional on ResultCard component

civ2boss:


metagrover (approved):

metagrover:
Thanks for contributing @civ2boss 💯

Issue #852, opened 2019-02-26
`defaultQuery` in MultiList component is not getting executed


lakhansamani:
**Issue Type:**
bug

**Platform:**
ReactiveSearch for Web

**Description:**
In Aggregation network request, `defaultQuery` prop value is not appended.

**Minimal reproduction of the problem with instructions:**
https://codesandbox.io/s/32koxx9n0p?fontsize=14

**Reactivesearch version:** 
`3.0.0-beta.4`

**Browser:**
all

cedricium:
Until this is fixed, is there any workaround for this?

For some background info, I use `defaultQuery` to filter out results by requiring them to match a specific id.

allendol:
Same problem exists on  SingleDropdownList.
Just go to https://codesandbox.io/s/32koxx9n0p?fontsize=14 and change the MultiList to SingleDropdownList
Same problem exist.

PR #853, opened 2019-02-26
fix(#852): defaultQuery bug in list components

bietkul:


packages/web/src/components/list/utils.js
```diff
@@ -34,7 +34,7 @@ const getAggsQuery = (query, props) => {
 };
 
 const getCompositeAggsQuery = (query, props, after) => {
-	const clonedQuery = { ...query };
+	const clonedQuery = { ...query, ...(props.defaultQuery && props.defaultQuery()) };
```
metagrover: Shouldn't we do this merging of queries at the end of these functions (right before we return the final query)? Then, it will ensure that if `aggs` is present in the defaultQuery, it will over-write the generated `aggs` query.

metagrover (changes requested):

packages/web/src/components/list/utils.js
metagrover: Also, we should only take `query` and `aggs` from `defaultQuery` here, since we apply other query-options on the root query [here](https://github.com/appbaseio/reactivesearch/blob/cf59fe76590d55ad432b541dc10bfed6f5b0c4fc/packages/web/src/components/list/MultiList.js#L383).

What do you think?

bietkul: cool, will update the PR

Issue #871, opened 2019-03-07
Non linear steps


ghost:
**Issue Type:**
Enhancement

**Platform:**
Web

**Description:**
I would like the ability to increment then steps of the slider in a non linear way. For example on a slider ranging from 1 to 1000. I want like 1 - 500 to take up 85% of the width of the slider because this is where most of the values are.

See an example here:
https://refreshless.com/nouislider/slider-values/#section-non-linear-step

metagrover:
Looks good! Open to accepting PRs on this.

ghost:
Will take a swing at this one.

Issue #831, opened 2019-02-14
DynamicRangeInput 


ghost:
**Issue Type:**
Feature request/question

**Platform:**
Web

**Description:**
I need the functionality of the DynamicRangeSlider: to show the highest and lowest values.
And the functionality of the RangeInput: Modify the highest and lowest values by entering a value into the input.

Is there a way that this can be done with the existing components or if not It would be awesome if there could be one.

adadiehl:
I've solved this problem by using an external elasticsearch API call to store the min/max values in state, then supply these as args to RangeInput's range prop. An internal solution would be better, but this might get you up and going in the mean time.

PR #878, opened 2019-03-13
web: Non linear stepping on slider components (#871) & DynamicRangeInput component (#831) & Scroll top bug fix for edge (#886)

ghost:
Expose Rheostat `algorithm` and `snapPoints` props on the RangeSlider and DynamicRangeSlider components with the desired result of allowing non linear stepping.

### DIFF
--- a/packages/web/src/components/range/DynamicRangeSlider.d.ts
+++ b/packages/web/src/components/range/DynamicRangeSlider.d.ts
@@ -17,7 +17,7 @@ export interface DynamicRangeSliderProps extends CommonProps {
 	rangeLabels?: (...args: any[]) => any;
 	react?: types.react;
 	showHistogram?: boolean;
-	snapPoints?: number[] | (start: number, end: number, stepValue: number) => number[];
+	snapPoints?: (start: number, end: number, stepValue: number) => number[];
 	algorithm?: {
 		getValue: (position: number, min: number max: number) => number;
 		getPosition: (position: number, min: number max: number) => number;
--- a/packages/web/src/components/range/DynamicRangeSlider.d.ts
+++ b/packages/web/src/components/range/DynamicRangeSlider.d.ts
@@ -17,7 +17,7 @@ export interface DynamicRangeSliderProps extends CommonProps {
 	rangeLabels?: (...args: any[]) => any;
 	react?: types.react;
 	showHistogram?: boolean;
-	snapPoints?: number[];
+	snapPoints?: number[] | (start: number, end: number, stepValue: number) => number[];
 	algorithm?: {
 		getValue: (position: number, min: number max: number) => number;
 		getPosition: (position: number, min: number max: number) => number;
--- a/packages/web/src/components/range/DynamicRangeSlider.d.ts
+++ b/packages/web/src/components/range/DynamicRangeSlider.d.ts
@@ -17,6 +17,11 @@ export interface DynamicRangeSliderProps extends CommonProps {
 	rangeLabels?: (...args: any[]) => any;
 	react?: types.react;
 	showHistogram?: boolean;
+	snapPoints?: number[];
+	algorithm?: {
+		getValue: (position: number, min: number max: number) => number;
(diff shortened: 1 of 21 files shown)


ghost:
Actually the dynamic range slider needs more work done on it. Can you just review what I have done so far and ensure that I am working within the contribution guidelines.

packages/web/src/components/range/DynamicRangeSlider.js
```diff
@@ -197,8 +197,17 @@ class DynamicRangeSlider extends Component {
 	};
 
 	getSnapPoints = () => {
-		let snapPoints = [];
 		let { stepValue } = this.props;
+
+		if (this.props.snapPoints) {
+			if (typeof this.props.snapPoints === 'function') {
```
metagrover: This doesn't go with your provided typedef - `snapPoints?: number[];`

metagrover (changes requested):

ghost:
@metagrover I have updated the type definitions and also added a new DynamicRangeSlider component. If you are happy with the changes I can also commit documentation.

packages/web/src/components/range/DynamicRangeSlider.d.ts
```diff
@@ -17,6 +17,11 @@ export interface DynamicRangeSliderProps extends CommonProps {
 	rangeLabels?: (...args: any[]) => any;
 	react?: types.react;
 	showHistogram?: boolean;
+	snapPoints?: number[] | (start: number, end: number, stepValue: number) => number[];
```
metagrover: I am wondering if we should just stick to one type here - it can always be a function which can return the static list of numbers if you wish to (irrespective of the given params). This way, it will be easier for anybody using this because now they always have to stick to one format and it will be easily maintainable from the development/docs standpoint. Thoughts?

metagrover (changes requested):

packages/web/src/components/range/DynamicRangeSlider.d.ts
ghost: Agreed. This makes more sense. Probably unlikely that anyone would use an array literal here anyway and even if they do want to they still can.

jyash97 (changes requested):
@george-dotdev Awesome work on the component. 👍 
Can we also handle the input values validation ? For example, I can put the start date more than that of the end date. Check this screenshot:
image

ghost:
@metagrover @jyash97 Have added validation as requested.  I added the validation to the existing RangeInput component too which didn't have any validation yet. I added the validation to the `blur` event of the input field so that the input isn't updated before the user has finished typing. We may also want to consider the validating when the input goes below `0`. 

I also added a new `inputs` property to the DynamicRangeInput component which allows better control over the input field rendering. This allows users to do things like correctly format money input fields using other components or use `text` input fields.
```
inputs={({ innerRef, value, inputFormat }) => {
	return (
		<div className="flex -margin-top-2x">
			<div className="grow">
				<input className="no-input tertiary-link" value={inputFormat(value.start)} ref={innerRef.start} readOnly/>
			</div>
			<div className="grow flex h-end">
				<input className="no-input tertiary-link text-right" value={inputFormat(value.end)} ref={innerRef.end} readOnly/>
			</div>
		</div>
	)
}}
```

Added `prefix` and `suffix` inputs too which add text before and after the inputs fields to give the user additional context on the value. note the `$` in the below example. 
image

Added new `inputFormat` and `inputUnformat` props which allow manipulation of the input field values. In the following example I am using it to add a `%` inside of the input field.

```
inputFormat={(value) => {
  return `${value}%`
}}
inputUnformat={(value) => {
  return value.replace('%', '')
}}
```
Screen Recording 2019-03-22 at 4 42 53 pm

Also made it so that the input fields update as the user drags the slider. This makes for better UX as you can see the exact value as you drag.

Screen Recording 2019-03-22 at 4 33 40 pm (2)

ghost:
Any changes you want me to make @bietkul ?

PR #882, opened 2019-03-15
Add parseDate prop

jyash97:
Sync #617


metagrover (changes requested):
We should also add these in typescript definitions

jyash97:
Done 🏁

metagrover (approved):

Issue #843, opened 2019-02-19
RFC: Better specs for render props for result components


metagrover:
**Issue Type:**
RFC

**Platform:**
All

**Description:**
With our current spec, we have introduced a set of unintended opinions into the component rendering control. Given how `loader` prop works, it is hard for a user to take control of loader rendering. For instance, some users may want to hide the results when the results are loading, or some may not. Again, the loader behavior is very ambiguous when it comes to `pagination` and infinite loading setup.

I strongly feel that we should improve our spec to merge error, loading and result in one render prop. We can keep separate event callbacks via `onErrors`, `onData`, etc, but for render props, it would make a lot more sense if these things are passed to one single entity to take control of the rendering and providing developers with more flexibility and no rendering opinions from the library:

```js
renderAllData={({ loading, error, data }) => (...)}
```

We will need to think of a way to enable this spec (or) an equivalent behavior with `renderData` which runs for individual items in the results set.

lakhansamani:
Along, with the above specs, we can also add `stats` and make pagination and infinite loading optional props, this can help developers to implement their own pagination component or scroll logic, thus allowing them to customize ui completely

metagrover:
Absolutely, that sounds good. This will further improve the rendering control.

lakhansamani:
Also, we should consider using [Render Prop](https://reactjs.org/docs/render-props.html) that way developer can use this data inside `ReactiveList` children

```
<ReactiveList ... >
  { ({ loading, error, data }) => (
    <div> ... </div>
  )}
</ReactiveList>
```

bietkul:
Similar to the Query component from Apollo Client
https://www.apollographql.com/docs/react/essentials/queries.html#basic

jyash97:
We should also do this in DataSearch and CategorySearch. Consider this case: 
If we use renderAllSuggestions, we want loader when the suggestions are loading and an empty state when there are no suggestions. In the current situation, I see both loader and empty state.

Preview: 
![image](https://user-images.githubusercontent.com/22376783/53282832-5cd0f980-3763-11e9-8ba1-ec437eb99060.png)

metagrover:
Thoughts on this:

```js
<ReactiveList>
    {({ data, error, loading, ... }) => (
        <ResultsWrapper>
            {
                data.map(item => (
                    <ResultCard>
                        <ResultCard.Image></ResultCard.Image>
                        <ResultCard.Title></ResultCard.Title>
[...]

PR #895, opened 2019-03-26
feat(web): Better specs for render props for result/search/list components (#843)

bietkul:


packages/web/src/components/list/MultiDataList.js
```diff
@@ -364,28 +364,53 @@ class MultiDataList extends Component {
 
 	handleClick = (e) => {
 		const { value, onChange } = this.props;
-		const { value: listValue } = e.target;
-		if (value === undefined) {
-			this.setValue(listValue);
-		} else if (onChange) {
-			onChange(listValue);
+		if (isEvent(e)) {
```
metagrover: what's the purpose of this? This `if` condition adds to an already complicated logic flow. Can you please elaborate on why have we added this and briefly go over the updated flow for controlled/uncontrolled components?

packages/web/src/components/list/MultiDataList.js
```diff
@@ -364,28 +364,53 @@ class MultiDataList extends Component {
...
 			return true;
 		});
+		return listItems;
+	}
+
+	render() {
+		const { selectAllLabel, showCount, renderItem } = this.props;
+		const { options } = this.state;
+
+		if (!this.hasCustomRenderer && options.length === 0) {
```
metagrover, line 406: This should be `||` (OR) imo

metagrover (changes requested):

packages/web/src/components/list/MultiDataList.js
bietkul: `&&` is intentional, User should has control to decide that what to do with empty results when custom renderer is being used.

metagrover: got it! 👍

packages/web/src/components/list/MultiDropdownList.js
```diff
@@ -386,15 +385,38 @@ class MultiDropdownList extends Component {
...
-			onChange(item);
+		if (isEvent(e)) {
+			const { value: listValue } = e.target;
+			if (value === undefined) {
+				this.setValue(listValue);
+			} else if (onChange) {
+				onChange(listValue);
+			}
+		} else {
+			this.setValue(e);
```
metagrover: So I get the `isEvent` logic here is used to include user's integration for the click handling. I wonder what is the reason for adding `this.setValue(e);` here in the else block?

packages/web/src/components/result/ReactiveList.d.ts
```diff
@@ -8,20 +8,22 @@ export interface ReactiveListProps extends CommonProps {
...
 	dataField: string;
 	defaultQuery?: (...args: any[]) => any;
 	defaultSortOption?: string;
 	excludeFields?: Array<string>;
 	innerClass?: types.style;
+	infiniteScroll?: boolean;
 	includeFields?: Array<string>;
 	loader?: types.title;
+	render?: (data: any) => any;
 	renderAllData?: (data: any) => any;
```
metagrover: Are we not removing `renderAllData`?

packages/web/src/components/result/ReactiveList.js
```diff
@@ -27,9 +27,17 @@ import PoweredBy from './addons/PoweredBy';
 
 import Flex from '../../styles/Flex';
 import { resultStats, sortOptions } from '../../styles/results';
-import { connect, isFunction } from '../../utils';
+import { container } from '../../styles/Card';
+import { container as listContainer } from '../../styles/ListItem';
+import { connect, isFunction, getComponent, hasCustomRenderer } from '../../utils';
 
 class ReactiveList extends Component {
+	static ResultsCardWrapper = ({ children, ...rest }) => (
```
metagrover: It should be `ResultCardsWrapper`

packages/web/src/components/result/ReactiveList.js
```diff
@@ -27,9 +27,17 @@ import PoweredBy from './addons/PoweredBy';
...
+		<div className={container} {...rest}>{children}</div>
+	);
+	static ResultsListWrapper = ({ children, ...rest }) => (
```
metagrover: It should be `ResultListWrapper`

packages/web/src/components/result/ReactiveList.js
```diff
@@ -119,7 +127,7 @@ class ReactiveList extends Component {
 		this.setReact(this.props);
 
 		this.domNode = window;
-		if (!this.props.pagination) {
+		if (this.showInfiniteScroll) {
```
metagrover, line 130: Infinite scroll doesn't have an entity to render - so not sure if this prop name makes sense? I think we can remove `show` from here.

packages/web/src/components/list/MultiDataList.js
bietkul: It's to detect that whether the passed argument is a synthetic event or not, if not then it'll treat the passed argument as value.
I'm exposing the `handleChange` method in `render` prop, there might be a case where user wants to change the value by `onClick` event rather than any `radio` or `checkbox` input, this condition handles it well.

packages/web/src/components/result/ResultList.js
```diff
@@ -1,84 +1,50 @@
...
 
 	render() {
-		const { renderData, ...props } = this.props;
-
-		return <ReactiveList {...props} renderData={this.renderAsListItem} listClass={container} />;
+		const {
+			children, href, small, target, ...props
+		} = this.props;
+		let hasImage = false;
+		React.Children.forEach(this.props.children, (o) => {
```
metagrover, line 19: This can some perf issues imo - we iterate over all the elements for every item in the result-list, on every render. Ideally, you should move this code in the lifecycle method and use state to provide `hasImage`?

Or, don't compute `hasImage` here at all. Compute it in the ListItem?

metagrover (changes requested):

packages/web/src/components/list/MultiDropdownList.js
bietkul: Think for one use case in which I want to show a button instead of radio input in my custom renderer, on clicking of which I want to update the current selected value.
for e.g

```
{ 
 ({ data, handlechange }) => data.map(item => (<div onClick={() => handleChange(item.label)}>{item.label}</div>))
}
```

packages/web/src/components/result/ReactiveList.d.ts
bietkul: thanks, I'll fix it

packages/web/src/components/result/ReactiveList.js
bietkul: ok

packages/web/src/components/result/ReactiveList.js
bietkul: ok

packages/web/src/components/result/ResultList.js
bietkul: cool, `state` can be a better option.

packages/web/src/components/result/ReactiveList.js
bietkul: Actually, It is not a prop, it is a computed value.

packages/web/src/components/list/MultiDataList.js
lakhansamani: I feel e should always be part of the parameter, it allows the user to manipulate things based on an event object. Second args can be value, WDYT?

bietkul: No need of the second parameter since now we're handling both cases with only one param.
If you pass an event then the value will be extracted from it otherwise the parameter will be treated as a supplied value.

metagrover: That would certainly be a better spec. We can do it, but it's going to be a massive change, and we don't keep track of the events after the entry funnel of these callbacks. At times, these callbacks get triggered via changes like SelectedFilters clearing.

packages/web/src/components/list/MultiDropdownList.js
lakhansamani: I think if `e` passed in handleChange is value instead of an event, then it would be used for setting that value, but it should be `this.setValue(value)` right?

bietkul: Right, but `this.setValue(e)` is fine since we have to set the value which has been provided by user.

lakhansamani: This might change how controlled components operate.

bietkul: Not sure, but that's the same thing we're doing while rendering the list items inputs, on changing of which we call the `handleChange` function.
Exposing `handleChange` in render props provides the easiest way to update the values instead of managing own state.

metagrover: how does the handleChange work with the controlled component behavior? I think you're now bypassing it and directly updating the component's value ~ uncontrolled. How will it work in a controlled setting?

lakhansamani: I agree we should not violate common ground rules for input state management.

bietkul: If some user is using the controlled setting then there is no meaning to use `handleChange` method.
I just want to simplify the update process if someone is using a custom renderer so a user can easily update the current value without managing their own state.

metagrover: so we can't use custom-rendering and controlled behavior together?

bietkul: Yes, we can.
then you have to maintain state yourself and define your custom state changer function.

bietkul: I don't think exposing `handleChange` will break any React paradigm.
It's just for the ease of users so one can easily bind their inputs to the component's state.

metagrover: Okay. That makes sense. Thoughts @lakhansamani ?

lakhansamani: Don't we already have one such method for a user to manage state, i.e. `onChange` then why to expose `handleChange`?

metagrover: onChange won't get triggered here. It only works with controlled behavior - which is not handled via the handleChange method in the render prop here.

lakhansamani: Okay, then it makes sense, though this method can add more confusions to api.

metagrover: I agree. I think we should support the controlled behavior here as well - for consistency and ease of use. Here's how I'm thinking of it:

```js
handleChange = (e) => {
    let currentValue = e;
	if (isEvent(e)) { currentValue = e.target.value }

   // handle the behavior as earlier
}
```

lakhansamani: Yes, this looks good 👍

bietkul: Perfect, I'll do the changes.

metagrover (approved):

PR #898, opened 2019-03-28
Fix bug with sort returning to default

robert-armstrong:
sortOptionIndex was never update in handleSortChange causing the sort to return to the default in cases where query options were rebuilt in componentDidUpdate.


siddharthlatest:
Thanks for the PR @robert-armstrong! 🎉  Curious to know how you are using ReactiveSearch (if you don't mind to share), and if you have any other feedback for us.

metagrover (approved):

metagrover:
Thanks for the contribution 🎉 

We will roll this out in the upcoming release this week.

PR #901, opened 2019-04-01
fix: display SelectedFilter with custom URL params

### LINKED ISSUE #899: SelectedFilters stops displaying filters when using custom URL parameters
**Issue Type:**
bug

**Platform:**
web

**Description:**
I encountered this while debugging an issue w/ reactivesearch combined w/ react-router and additional custom URL params for other application state. When a URL change occurs on my custom parameters and URLParamsProvider attempts to resynchronize state, SelectedFilters drops any existing filters it had in place. While debugging, I can tell that this.props.selectedValues in URLParamsProviders has the right elements, but the showFilter value has becomes null. I believe this happens because of this:

https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/basic/URLParamsProvider.js#L27

I think you can reproduce this by using the https://opensource.appbase.io/reactivesearch/demos/producthunt demo. Add a filter and then execute this in DevTools:			

`
window.dispatchEvent(new Event('popstate'));
`

which is the same thing as called from

https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/basic/URLParamsProvider.js#L124.

You should see the filters disappear.

If you can confirm my logic here and point me in the right direction, I'm happy to generate a PR for a fix.

lakhansamani:
resolves #899

### DIFF
--- a/packages/web/src/components/basic/URLParamsProvider.js
+++ b/packages/web/src/components/basic/URLParamsProvider.js
@@ -24,7 +24,9 @@ class URLParamsProvider extends Component {
 			// update active components in selectedValues
 			Array.from(this.params.entries()).forEach((item) => {
 				try {
-					this.props.setValue(item[0], JSON.parse(item[1]));
+					const [component, value] = item;
+					const { label, showFilter, URLParams } = this.props.selectedValues[component];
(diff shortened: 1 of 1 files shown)


metagrover (approved):
💯

PR #905, opened 2019-04-02
RFC: Add showDimmedStars in RatingsFilter

jyash97:
RFC:

To remove grey stars from `RatingsFilter`. Not sure if this is good.

Thoughts ?


lakhansamani:
Any use case where we need to remove `grey` stars?

jyash97:
Thinking on this again. I think we can achieve this by changing the styles of the dimmed icon. Also, I cannot think of a use case for now and this is easily achievable so let's close this for now.

PR #906, opened 2019-04-02
feat(web): Added render prop in ReactiveComponent

bietkul:
- Change HOC based implementation to render prop API
- Renamed `onAllData` to `onData`, now it'll return an object with the following properties:
```
{
   data // parsed hits
   rawData // original hits
   aggregations // object
}
```


metagrover:
I think we will also need `selectedValues` from redux store here? Maybe filter it by the componentId and pass the value? What do you think @bietkul?

bietkul:
> I think we will also need `selectedValues` from redux store here? Maybe filter it by the componentId and pass the value? What do you think @bietkul?

I'm already providing it in `render` function, are you talking about the onData callback?
It's renamed to `value`

metagrover (approved):

PR #910, opened 2019-04-02
refactor(web): Remove `small` prop form ResultList Component

bietkul:


packages/web/src/components/result/ResultList.js
```diff
@@ -12,31 +12,35 @@ class ResultList extends Component {
 
 	state = {
 		hasImage: false,
+		isSmall: false,
 	};
 
 	componentDidMount() {
 		let hasImage = false;
+		let isSmall = false;
 		React.Children.forEach(this.props.children, (o) => {
```
metagrover: Instead of using `forEach` here, you should use methods like `every` or `some` - which don't run till loop completion. It will make the code more readable as then you can remove `!hasImage` from the `if` condition.

metagrover (changes requested):
Minor changes in the related code. Rest everything looks good 👌

packages/web/src/components/result/ResultList.js
bietkul: Updated the PR.
Please take a look now.

metagrover (approved):

PR #913, opened 2019-04-03
chore(maps): Upgrade Lifecycle for GeoDistanceDropdown

jyash97:
Working Demo
GDD

### DIFF
--- a/packages/maps/src/components/basic/GeoDistanceDropdown.js
+++ b/packages/maps/src/components/basic/GeoDistanceDropdown.js
@@ -46,41 +46,38 @@ class GeoDistanceDropdown extends Component {
 		let currentLocation = null;
 		let currentDistance = 0;
 
+		props.setQueryListener(props.componentId, props.onQueryChange, null);
+		props.addComponent(this.props.componentId);
+		this.setReact(props);
+
 		if (props.value) {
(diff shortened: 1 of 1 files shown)


lakhansamani (approved):

packages/maps/src/components/basic/GeoDistanceDropdown.js
```diff
@@ -49,74 +42,104 @@ class GeoDistanceDropdown extends Component {
...
-			});
+			currentDistance = selected.distance;
+		} else if (props.selectedValue) {
+			currentLocation = props.selectedValue.location;
+			const selected = props.data.find(
+				item => item.label === props.selectedValue.label,
+			);
+
+			currentDistance = selected.distance;
+		} else if (props.value) {
```
lakhansamani: this condition is same as `if` condition so it will never execute this block

lakhansamani (changes requested):

packages/maps/src/components/basic/GeoDistanceDropdown.js
jyash97: Thanks will fix this

lakhansamani (approved):

PR #914, opened 2019-04-03
chore(maps): Upgrade lifecycle for GeoDistanceSlider

jyash97:
GDS

### DIFF
--- a/packages/maps/src/components/basic/GeoDistanceSlider.js
+++ b/packages/maps/src/components/basic/GeoDistanceSlider.js
@@ -37,13 +37,6 @@ class GeoDistanceSlider extends Component {
 	constructor(props) {
 		super(props);
 
-		this.state = {
-			currentLocation: null,
-			currentDistance: props.range.start,
-			userLocation: null,
-			suggestions: [],
-			isOpen: false,
-		};
 		this.type = 'geo_distance';
 		this.locked = false;
 		this.coordinates = null;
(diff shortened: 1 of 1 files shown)


packages/maps/src/components/basic/GeoDistanceSlider.js
```diff
@@ -52,71 +45,68 @@ class GeoDistanceSlider extends Component {
...
-				},
-			);
+		if (props.selectedValue && props.selectedValue.location) {
+			currentLocation = props.selectedValue.location;
+			this.getCoordinates(props.selectedValue.location, () => {
+				if (props.selectedValue.distance) {
+					this.setDistance(props.selectedValue.distance);
+				}
+			});
+		} else if (props.defaultSelected && props.defaultSelected.location) {
```
metagrover: we should remove `defaultSelected` and use `value` - `defaultValue` props.

packages/maps/src/components/basic/GeoDistanceSlider.js
```diff
@@ -52,71 +45,68 @@ class GeoDistanceSlider extends Component {
...
+	componentDidUpdate(prevProps) {
+		checkPropChange(this.props.react, prevProps.react, () => this.setReact(this.props));
 
-		checkSomePropChange(this.props, nextProps, ['dataField', 'nestedField'], () => {
-			this.updateQuery(this.state.currentDistance, nextProps);
+		checkSomePropChange(this.props, prevProps, ['dataField', 'nestedField'], () => {
+			this.updateQuery(this.state.currentDistance, this.props);
 		});
 
 		if (
```
metagrover, line 91: same here - watch for changes on `value` prop if defined.

metagrover (changes requested):
We need to cover controlled/uncontrolled component use-case as well.

packages/maps/src/components/basic/GeoDistanceSlider.js
jyash97: Have done defaultValue changes but wasnt able to make it control.

jyash97:
I have done this changes in #918

PR #919, opened 2019-04-04
fix(web): continue parsing URL params even if one fails

robert-armstrong:
Similiar to code in #901, only fail on a single param, not the entire Array. This can happen if you have a custom param before or in the middle of reactivesearch filter params. This could also validate on a list of currently configured filter names, but I didn't see an easy way to add that.


lakhansamani (approved):
This is perfect 💯

PR #920, opened 2019-04-05
feat(web): Update demos and examples

bietkul:


packages/web/examples/DataSearch/src/index.js
```diff
@@ -1,79 +1,95 @@
...
+					componentId="SearchResult"
+					dataField="original_title"
+					size={10}
+					className="result-list-container"
+					pagination
+					react={{
+						and: 'BookSensor',
+					}}
+					render={({ data }) => (
+						<ReactiveList.ResultCardsWrapper>
```
jyash97, line 39: Shouldn't it be `ResultCardWrapper`?

bietkul: It's ok, we are using `ResultCardsWrapper`.

PR #921, opened 2019-04-05
 fix(native-site): update example links

lakhansamani:


bietkul (approved):

siddharthlatest:
@lakhansamani Are we missing the build step here? The website isn't still updated.

lakhansamani:
Shouldn't that be automated for sites?

siddharthlatest:
We use gh-pages for these sites, and it's not automated here currently.

lakhansamani:
Okay I checked code, we are using following scrip `deploy-site` to deploy website, so let me complete that step

Issue #926, opened 2019-04-05
App crash on IE11 due to popstate event


willopez:
**Issue Type:**
Bug

**Platform:**

web V3-beta8

**Description:**
On IE11, ReactiveSearch crashes when changing from one page to another. The root cause is the following command: 
```
window.dispatchEvent(new Event('popstate'));
```
on IE11 it causes the following error: 
```
Object does't support this action
```
ReactiveSearch is using the `popstate` event here:

https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/basic/URLParamsProvider.js#L126

**Screenshots:**

![ei11](https://user-images.githubusercontent.com/486340/55651126-7fabf000-579c-11e9-8a04-36f557c843e1.png)

**Minimal reproduction of the problem with instructions:**

**Reactivesearch version: v3-beta.8

**Browser:** IE11

jyash97:
Hey @willopez 
Thanks for filing this. It looks like `new Event` is not supported in `IE11`. I will take a look at this.

willopez:
Hi @jyash97 thanks for looking at this issue. I believe adding this will fix the issue: 
```
let event;
if (typeof(Event) === 'function') {
    event = new Event('popstate');
} else {
    event = document.createEvent('Event');
    event.initEvent('popstate', true, true);
}

window.dispatchEvent(event);
```

willopez:
@jyash97 I have created a PR to fix this issue: #928

jyash97:
Fixed [here](https://github.com/appbaseio/reactivecore/pull/35)

lakhansamani:
@willopez we added new release `3.0.0-rc.1` which includes above fix

willopez:
great! thanks

PR #928, opened 2019-04-08
fix: correctly fire popstate event on IE11 to prevent app crash.

willopez:
Fixes: #926 

Adds conditional logic to correctly fire `popstate` event on IE11.

### DIFF
--- a/packages/web/src/components/basic/URLParamsProvider.js
+++ b/packages/web/src/components/basic/URLParamsProvider.js
@@ -123,7 +123,16 @@ class URLParamsProvider extends Component {
 				: window.location.search;
 
 			if (searchString !== this.searchString) {
-				window.dispatchEvent(new Event('popstate'));
+				let event;
+				if (typeof (Event) === 'function') {
+					event = new Event('popstate');
+				} else {
+					// Correctly fire popstate event on IE11 to prevent app crash.
+					event = document.createEvent('Event');
+					event.initEvent('popstate', true, true);
+				}
+
+				window.dispatchEvent(event);
 			}
 		}
 	};


lakhansamani (commented):
Looks good 👍

lakhansamani (approved):

packages/web/src/components/basic/URLParamsProvider.js
```diff
@@ -123,7 +123,16 @@ class URLParamsProvider extends Component {
 				: window.location.search;
 
 			if (searchString !== this.searchString) {
-				window.dispatchEvent(new Event('popstate'));
+				let event;
+				if (typeof (Event) === 'function') {
```
jyash97, line 127: We can also move this logic to polyfill file so that we can have all the polyfills at one place. Thoughts @lakhansamani ?

willopez:
Thank you for merging this @lakhansamani. Can a new package version be published to npm, so I can integrate this update into my app. thanks.

packages/web/src/components/basic/URLParamsProvider.js
lakhansamani: @jyash97 true, I was not aware of `src/polyfill.js`.

Let me revert this commit and fix it 👍

lakhansamani: @jyash97 I agree, I was not aware of `polyfill` in `reactivecore`.

Have added PR for same, https://github.com/appbaseio/reactivecore/pull/35
Please check that once. 
If it works well, I will revert this PR.

lakhansamani:
@willopez as per above discussion, once the polyfill is added to our core library, will release new version on npm.

willopez:
got it, thank you. And looking forward to the next RC release!

PR #930, opened 2019-04-09
feat(maps): Add error handling in ReactiveMap

bietkul:


packages/maps/src/components/result/ReactiveMap.js
```diff
@@ -809,7 +822,9 @@ ReactiveMap.propTypes = {
 	markers: types.children,
 	renderAllData: types.func,
 	renderData: types.func,
+	renderError: types.title,
```
lakhansamani, line 825: We can convert this into `function`

packages/maps/src/components/result/ReactiveMap.js
```diff
@@ -865,6 +880,7 @@ const mapStateToProps = (state, props) => ({
 			&& state.selectedValues[`${props.componentId}-page`].value - 1)
 		|| 0,
 	time: (state.hits[props.componentId] && state.hits[props.componentId].time) || 0,
+	error: state.error[props.componentId],
```
lakhansamani, line 883: We should have setError from `actions` to set Errors in case GoogleMap / OpenStreetMap throws any error.

packages/maps/src/components/result/ReactiveMap.js
```diff
@@ -89,7 +90,7 @@ class ReactiveMap extends Component {
 		};
 
 		this.internalComponent = `${props.componentId}__internal`;
-		props.setQueryListener(props.componentId, props.onQueryChange, null);
+		props.setQueryListener(props.componentId, props.onQueryChange, props.onError);
```
lakhansamani, line 93: Here we can call `setError` and set the error for this component in redux store

lakhansamani (changes requested):

packages/maps/src/components/result/ReactiveMap.js
bietkul: Didn't get you, we're already maintaining error state in redux.
`onError` is a callback which gets triggered when there is an error in `_msearch` request.

packages/maps/src/components/result/ReactiveMap.js
bietkul: It can also be a string.

packages/maps/src/components/result/ReactiveMap.js
lakhansamani: Okay, but for maps we might have error other than elasticsearch error, which we can add in redux store, WDYT?

packages/maps/src/components/result/ReactiveMap.js
bietkul: @lakhansamani Is there any way to catch errors from GoogleMap / OpenStreetMap?
i.e are we already doing it somewhere?

packages/maps/src/components/result/ReactiveMap.js
bietkul: Right but it's not related to this line of code.

packages/maps/src/components/result/ReactiveMap.js
lakhansamani: For google map one check we can add is

https://github.com/appbaseio/reactivesearch/pull/930/files#diff-336a5c69e66e07ecdcb03bdcb94c0868R272

bietkul: Actually, that check is it to avoid error in case of Network failure, if there is no connection then the maps script won't be loaded and `window.google` will be absent.
Before, It was crashing the app.
In case of network failure, _msearch will also fail so `renderError` will work.
Although we can handle maps error in `GeoDistanceDropdown` but in `ReactiveMap` there is no use case for now.

lakhansamani (approved):

PR #932, opened 2019-04-09
fix(maps): prop warnings + reactivemap export

lakhansamani:


packages/maps/src/components/result/ReactiveGoogleMap.js
```diff
@@ -354,17 +354,17 @@ class ReactiveGoogleMap extends Component {
...
-	addComponent: types.funcRequired,
-	setMapData: types.funcRequired,
-	loadMore: types.funcRequired,
-	removeComponent: types.funcRequired,
-	setQueryListener: types.funcRequired,
+	addComponent: types.func,
+	setMapData: types.func,
+	loadMore: types.func,
+	removeComponent: types.func,
+	setQueryListener: types.func,
```
bietkul: I have one question, what's the need to define these props here.
Since we're not using it here, we can simply remove it.

bietkul (changes requested):

packages/maps/src/components/result/ReactiveGoogleMap.js
lakhansamani: True, let me just add props that are part of api, earlier it was just replica of `ReactiveMaps`

packages/maps/src/components/result/ReactiveMap.js
```diff
@@ -784,17 +784,17 @@ class ReactiveMap extends Component {
...
+	setQueryListener: types.func,
 	onQueryChange: types.func,
 	setPageURL: types.func,
-	setQueryOptions: types.funcRequired,
+	setQueryOptions: types.func,
 	setStreaming: types.func,
-	updateQuery: types.funcRequired,
-	watchComponent: types.funcRequired,
+	updateQuery: types.func,
+	watchComponent: types.func,
```
bietkul: It's okay to have `required` here.

packages/maps/src/components/result/ReactiveOpenStreetMap.js
```diff
@@ -191,9 +191,34 @@ class ReactiveOpenStreetMap extends Component {
...
+	defaultCenter: types.location,
+	defaultZoom: types.number,
+	defaultRadius: types.number,
+	defaultPin: types.string,
+	innerClass: types.style,
+	markers: types.children,
+	showMarkers: types.bool,
 	tileServer: types.string,
+	title: types.string,
+	error: types.title,
```
bietkul: remove `error` instead add `onError`

bietkul (approved):

packages/maps/src/components/result/ReactiveOpenStreetMap.js
lakhansamani: Okay

bietkul (approved):

PR #934, opened 2019-04-10
chore: Update Demos

jyash97:


site/demos/airbeds/webpack.config.js
```diff
@@ -1,7 +1,6 @@
 const path = require('path');
 
 module.exports = {
-	mode: 'development',
```
lakhansamani, line 4: Instead of changing mode for each demo and example, we can fix webpack version in site, which actually throws error regarding mode key
https://github.com/appbaseio/reactivesearch/blob/next/site/package.json#L35

Just make sure site build and start are not failing

jyash97: Done the changes https://github.com/appbaseio/reactivesearch/pull/936

PR #942, opened 2019-04-15
feat(web): Add script to update workspace  RS version after release

jyash97:


packages/web/package.json
```diff
@@ -22,7 +22,8 @@
 		"prepublish": "yarn run build",
 		"test": "jest --watch",
 		"typescript": "tsc -p tsconfig.json",
-		"build-example": "scripts/build-example.sh 3.0.0-rc.0"
+		"build-example": "scripts/build-example.sh 3.0.0-rc.3",
```
bietkul, line 25: I think there should be some way to pick the version from version field of `package.json`.

bietkul (changes requested):

packages/web/package.json
jyash97: Yeah that will be great, any idea or lead how this can be done ? 
I searched but didnt got a good way to do this.

bietkul: Check this https://github.com/kentcdodds/nps
You can define the scripts outside of the package.json
Ref. https://github.com/appbaseio/appbase-js

jyash97: Fixed in #959

Issue #900, opened 2019-03-29
`UpdateQuery` when `CustomQuery` changes


nadrane:
**Issue Type:**
Enhancement

**Platform:**
web

**Description:**
I've run into quite a few scenarios now where I have a Reactivesearch component that is is wrapped in an outer component that maintains its own state. The Reactivesearch's custom query is driven based on the state inside the outer component.

The problem with custom queries today is that the query does not update/execute when the custom query changes; the query only updates/executes when the `value` prop changes. But what should I do when the custom query is driven by both the value and some piece of state or prop maintained in an outer component?

Today, inside the `componentDidUpdate` lifecycle hook of each (many?) Reactivesearch component, we check if the `value` prop has changed, and if it has, then we update the query. I'm proposing that in cases where a custom query is used, we also check to see if the `customQuery` has deeply changed between re-renders and if it has, update the query. This behavior change would allow Reactivesearch components to behave far more flexibly.

What are people's thoughts? I'd be happy to implement this.

metagrover:
We used to have this functionality (in v2)

@bietkul -  can you please check on this? I think it stopped working after we changed the spec for customQuery prop.

Issue #867, opened 2019-03-05
Pre-fill components with defaultQuery


alexandre033:
Hi :),
i just need to know if it's possible to pre-fill the component compared to a reactiveList defaultQuery ? I save the query "{preference: ReactiveList}" so when i call this query the results items is perfect, it fits well with the selected filters, but the range component and multilist component are not up to date with the reactive list render. There is a way to save the query, reuse it and benefit from pre-filled fields (multilist, date..) ? Thanks a lot

Issue #531, opened 2018-09-21
Aggregations In ReactiveComponents and DataSearch


mb-3000:
**Issue Type:**
Trying to add aggregations into my master search box which is a DataSearch should be easier to do.

**Platform:**
Web

**Description:**
Ok so I would like to add some aggregations to a query so I can have a component somewhere else use those values for min/max values on a slider (so a date range thats draggable for instance)

I had a look at the comments in here and other stuff like ReactiveComponent and its ability to add `aggs` to a `defaultQuery` object and it kind of works.

I don't want to have to write by hand a DataSearch component again so I tried to embed it inside the `ReactiveComponent` tag but a peculiar thing is happening.

The DataSearch is using the defaultQuery from the outer ReactiveComponent and appending it into the query object.

So the code like this:
(I've abbreviated it a bit....)
```
const searchFields = [
  'collectionName',
  'collectionName.keyword',
  'imageName',
  'imageName.keyword'
];

const fieldWeights = [
  2,
  2
];

const highlightFields = [
  "imageName",
  "collectionName"
];

/**
 * Add aggregations to the end of the queries for min/max dates
 * @param value
 * @param props
 */
const aggOptions = () => ({
    aggs: {
      min_date: {
        min: {
          field: "dateCreated"
        }
      },
      max_date: {
        max: {
          field: "dateCreated"
        }
      }
    }
  }
);
// then in render()
...
          <ReactiveComponent
            componentId={'search'}
            defaultQuery={aggOptions}
          >
            <DataSearch
              componentId="searchTextSensor"
              showIcon={false}
              autosuggest={true}
              debounce={5}
              URLParams={this.props.useUrlParams}
              dataField={searchFields}
              fieldWeights={fieldWeights}
              highlight={true}
              highlightField={highlightFields}

            />
          </ReactiveComponent>
... 
```
Produces Json Like this:
(I've abbreviated it a bit....)
```
{
  "query": {
    "bool": {
      "must": [
        {
          "bool": {
            "must": {
              "bool": {
                "must": [
                  {
                    "bool": {
                      "should": [
                        {
                          "multi_match": {
                            "query": "man",
                            "fields": ["field^2"],
                            "type": "best_fields",
                            "operator": "or",
                            "fuzziness": 0
                          }
                        }, {
                          "multi_match": {
                            "query": "man",
                            "fields": ["collectionName^2"],
                            "type": "phrase_prefix",
                            "operator": "or"
                          }
                        }
                      ],
[...]

PR #954, opened 2019-04-19
fix: default query usage

### LINKED ISSUE #855: defaultQuery support in search component
**Issue Type:**
Enhancement

**Platform:**
All

**Description:**
This will enable customising only the suggestions query, and not the component's search query. With our current behavior, we use the same query to fetch suggestions and same query to fire the search query after the suggestion is selected - which can be updated via `customQuery`.

This will essentially enable users to customise both the queries separately.

bietkul:
This PR contains the following things
- Apply the `query` option from `defaultQuery` in search & list components
- Add `defaultQuery` support in search components
- Make `defaultQuery` and `customQuery` prop reactive i.e prop change will trigger a new query.

Fixes #855, #867, #852, #531, #900


packages/web/src/components/list/MultiList.js
```diff
@@ -365,6 +378,29 @@ class MultiList extends Component {
 		});
 	};
 
+	updateDefaultQuery = (queryOptions) => {
```
lakhansamani, line 381: Seems like repeating coding, can we make common util function for `updateDefaultQuery`?

packages/web/src/utils/index.js
```diff
@@ -55,3 +56,18 @@ export const hasCustomRenderer = (props = {}) => {
...
 export const isEvent = candidate =>
 	!!(candidate && candidate.stopPropagation && candidate.preventDefault);
+/**
+ * To check if two functions are identical
+ */
+export const isIdentical = (a, b) => {
+	if (a === b) {
+		return true;
+	}
+	if (typeof a === 'function' && typeof b === 'function') {
```
lakhansamani, line 64: Won't the above condition already check type for functions, since we are using `===`. I guess it might never go into this condition, as it will return true from line 64?

lakhansamani (changes requested):

packages/web/src/utils/index.js
bietkul: The first condition is just to cover the case when both have `undefined` or `null` value.
The second condition compares(deeply) the returned value of functions to avoid the infinite loop when someone uses arrow functions.

lakhansamani: I see in that case we can make the first condition more clear by writing
```
if (!a && !b) return true
```

packages/web/src/components/list/MultiList.js
bietkul: cool, will do something.

packages/web/src/utils/index.js
bietkul: agree

packages/web/src/components/search/CategorySearch.js
```diff
@@ -395,34 +422,60 @@ class CategorySearch extends Component {
...
-			this.updateQuery(this.internalComponent, value, this.props);
+			this.updateDefaultQuery(value, this.props);
 		} else {
-			this.updateQuery(this.props.componentId, value, this.props);
+			this.updateQuery(value, this.props);
 		}
 	}, this.props.debounce);
 
-	updateQuery = (componentId, value, props, category) => {
+	updateDefaultQuery = (value, props, category = this.state.currentCategory) => {
```
lakhansamani, line 431: How about using `defaultInternalQuery` here with one more param, i.e. `aggsQuery`?

packages/web/src/components/search/DataSearch.js
```diff
@@ -343,22 +353,44 @@ class DataSearch extends Component {
...
-			this.updateQuery(this.internalComponent, value, this.props);
+			this.updateDefaultQuery(value, this.props);
 		} else {
-			this.updateQuery(this.props.componentId, value, this.props);
+			this.updateQuery(value, this.props);
 		}
 	}, this.props.debounce);
 
-	updateQuery = (componentId, value, props) => {
+	updateDefaultQuery = (value, props) => {
```
lakhansamani, line 362: Same here, we can use`updateInternalQuery`. To make function simpler we can add switch cases over there for component types.

packages/web/src/components/search/DataSearch.js
```diff
@@ -343,22 +353,44 @@ class DataSearch extends Component {
...
+		}
+		props.setQueryOptions(this.internalComponent, {
+			...this.queryOptions,
+			...defaultQueryOptions,
+		});
+		props.updateQuery({
+			componentId: this.internalComponent,
+			query,
+			value,
+			componentType: 'DATASEARCH',
```
lakhansamani: Can we make ENUM for all component type we have and use it? WDYT?

lakhansamani (changes requested):

packages/web/src/components/search/DataSearch.js
bietkul: Things are different with search components.

packages/web/src/components/search/DataSearch.js
bietkul: > Can we make ENUM for all component type we have and use it? WDYT?

At that time it may need some extra work since they have been defined in `reactivecore` so we have to export them and use it here.
I think it's ok to just use strings, we can do that code refactoring later at some time.

PR #959, opened 2019-04-23
 feat: use nps for scripts

jyash97:
In testing I was getting some versioning issues and all the packages are currently updated. But I checked with pwd to ensure we are in the right directory.


scripts/build-web-example.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+echo Upgrading examples RS version to $1
+
+var=`pwd`
+for d in packages/web/examples/*/ ; do
+    cd $var/$d
+    yarn add @appbaseio/reactivesearch@$1
```
bietkul: Why do we need to use yarn here, again it'll take a long time to execute this script?
Can't you just modify the package.json with the particular version?

bietkul (changes requested):

scripts/build-web-example.sh
jyash97: have removed `yarn` & now we are just changing package version.

scripts/update-web-workspace.sh
```diff
@@ -0,0 +1,5 @@
+echo Upgrading related Workspace to $1
+
+cd packages/playground && yarn add @appbaseio/reactivesearch@$1 && cd ../maps && yarn add @appbaseio/reactivesearch@$1 && cd ../..
```
bietkul: remove `yarn` from here too.
and you're not updating the demos inside `site` folder.
I think only one script can work with a recursive function
- Start from the root and traverse all the folder
- Check if the folder has package.json with reactivesearch dependency
- update it to the latest
- move forward and repeat it for all the nested folders until the whole directory is covered

bietkul (changes requested):

scripts/update-web-workspace.sh
jyash97: > and you're not updating the demos inside site folder.

For `site` we already have scripts & we need yarn there to update `yarn.lock` & create a build.

> remove yarn from here too.

We have yarn.lock file in playground as well. So we may need to use yarn their also.

lakhansamani (approved):

bietkul (approved):

bietkul:
@jyash97 Can you resolve the conflicts?

jyash97:
Done 🏁

PR #973, opened 2019-05-09
chore: add release drafter

lakhansamani:
### Setup

* For release drafter to work correctly, we need to make `next` branch as the default branch or need to merge all changes into `dev`.
* In order to categories commits into `feature`, `fix` and `chore` we need to add respective labels to PR before merging them.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,11 @@
+categories:
+    - title: 🚀 Features
+      label: feature
+    - title: 🐛 Bug Fixes
+      label: fix
+    - title: 🔧 Maintenance
+      label: chore
+template: |
+    ## Changes
+
+    $CHANGES


jyash97 (commented):
Is there any way we can differentiate `React`, `Vue` and `Native` changes? 

Also, we can add a `PR labeler` or `Probot` to automatically add the respective labels, wdyt ?

lakhansamani:
Yes, we can differentiate between React, Vue, Native for the release, but I think, that should be part of tagging, eg. `react-v3.0`, WDYT?

jyash97:
Not too sure, but I think it should work. What about probot for automatic tagging ?

lakhansamani:
Yes, will research on that and will try to integrate it.

jyash97:
We can integrate it later, for now release drafter makes more sense 👍

jyash97 (approved):

PR #983, opened 2019-05-15
Bug Fix: When the categoryField is not specified on the Category?Sear…

tapasvimoturu:
…ch component, auto suggest is broken. Given that the category search field is optional according to the documentation here

https://opensource.appbase.io/reactive-manual/search-components/categorysearch.html#props

This should have worked, but the code that recommends the final suggesiton list is only merging the two results when category search is enabled. This can be replicated in the demo example from the documentation site by removing the category field. I have created a codesandbox.io that showcases the broken search functionality here
https://codesandbox.io/embed/2orqz6rryn

The fix ensures the the parsedSuggestions getter always merges the categoriesResults and the suggestionsResults.

### DIFF
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -732,7 +732,7 @@ class CategorySearch extends Component {
 	get parsedSuggestions() {
 		let finalSuggestionsList = [];
 		let suggestionsList = [];
-
+        let categorySuggestions = [];
 		// filter out empty categories
 		const filteredCategories = this.filteredCategories;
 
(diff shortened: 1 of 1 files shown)


tapasvimoturu:
@bietkul  Any ETA on the review and provide feedback for changes?

bietkul:
@tapasvimoturu Thanks for creating the PR but I don't think this PR will make any changes to the functionality.

Please note that the auto search functionality won't work if you don't define the `categoryField` in `CatgorySearch`, You can use `DataSearch` component for your use case.

And btw the `_msearch` request throws an error when you don't define the `categoryField` (check the network response), ReactiveSearch wasn't reporting such errors before but we have fixed it in the latest version so now these errors can be handled by using `onError`.

PR #988, opened 2019-05-17
Improve DynamicRangeSlider in Vue

rubenmoya:
### Description
This PR improves the `DynamicRangeSlider` in the Vue version, we tried to integrate this component into our project and ran intro a couple of problems with the range/currentValue not updating as expected or breaking the component entirely.

The commits are pretty self explanatory, I tried to make them as atomic as possible, but basically the changes I made are:

- Update `vue-slider-component` to the latest version
- Use `this.xxx` instead of `this.$props` or `this.$data`
- Use `range__internal` like the React version
- Small refactors in the whole component

There are a couple of things I'm not sure about my changes, the use of `range__internal`, I've seen it in the React version and I think it is what fixed the problems we had with our range selector going crazy, but I'm not that familiar with the project yet to know for sure, the other one is the use of `NoSSR`, it seems like the latest version of `vue-slider-component` handles server side rendering, but I'd love if you could give me a hand checking it.

And that's it, any feedback is welcome since this is my first PR in the project and I've changed a lot of things.


jyash97:
Hey @rubenmoya 
Can you elaborate on the issues that you faced while using DynamicRangeSlider ?

rubenmoya:
Hello @jyash97, thank you for answering!

This is the problem:

[Gif removed, could add again if needed]

The initial values are `{ min: 3194, max: 26887 }`, after selecting `Vespa` it changes to `{ min: 3194, max: 5890 }`, if I remove the filter it keeps the same value, but adjusts the totals, in this case in my opinion, since the filter was at the ends of the range it shouldn't keep the same value (5890), if now I select 'Moto-Guzzi', since the values are `{ min: 9134, max: 26887 }` the range slider breaks because both, the min and max value are outside the totals.

When the value is at the end of the range we should update it when the totals update.

jyash97:
Thanks for the update @rubenmoya  
Will review the PR within this week and let you know if any changes are required.

jyash97:
Just need to confirm one thing @rubenmoya are you using `react` prop in the component, or is the changes happening without react prop ?

jyash97 (commented):
Hey @rubenmoya  thanks for the PR, functionality wise it works great. I have few questions:

1. Any reason for removing `VueSSR`
2. Also removing of `this.$props` , though with `this` it will work.

cc @bietkul

rubenmoya:
Hi!

Yes, I was using the `react` prop.

I removed `this.$props` because just using `this` works, and it's the recommended way to do it, since it's a Vue related project I think it's better to stick to the Vue guidelines.

About the `VueSSR` component, the latest version os `vue-slider-component` works with server side rendering ([more info](https://nightcatsama.github.io/vue-slider-component/#/?hash=server-side-rendering-ssr)), although I'm not sure how to make it work since it complained about not having any css loader.

mauxtin:
I have been using this for a few days now and can confirm everything works fine 🚀

Thanks @rubenmoya!

mauxtin:
Taking a closer look at this, I found three problems.

1. After using this PR and having `URLParams` set to `true`, the page automatically redirects on page load from `page.com/xyz` to `page.com/xyz?range=xxxx`
I rechecked with the latest beta and this only happens after this PR.

2. On page load with no query string on the url (i.e. `page.com/xyz`) I am now getting this error
`The prop "interval" is invalid, "(max - min)" cannot be divisible by "interval"`
The slider appears though.

3. On page load with slider query (i.e. `page.com/xyz?range=xxxx`) I am getting the following error
`[Vue warn]: Error in beforeMount hook: "TypeError: Cannot read property 'start' of null"`
which is caused [here](https://github.com/appbaseio/reactivesearch/pull/988/files#diff-9d4d2f754188bf82bf478fa4bb556231R139)
When this error occurs the slider does not appear at all.

bietkul:
Thanks, @mauxtin for the PR review.
@rubenmoya Can you please address the issues raised by @mauxtin?

jyash97:
Thanks 💯  @rubenmoya  for initial implementation we have taken forward and fixed URLParams also in #1189 .

Thank you @mauxtin  for reviewing the PR and making the library better 💯

PR #996, opened 2019-05-22
add exhaustive service options prop for geodistance

### LINKED ISSUE #991: Allow to pass options to AutocompleteService when using GeoDistanceSlider
**Issue Type:** Enhancement

**Platform:** Web

**Description:**
I'd like to be able to pass `window.google.maps.places.AutocompleteService` options, so that I'm able to limit the search to a certain scope or just modify the way how I get data from Google.

**Reactivesearch version:** 2.18.0

**Browser:** all

**Other:**

You can pass various options https://developers.google.com/maps/documentation/javascript/reference/places-autocomplete-service

I think one should be able to pass options dictionary as a prop to `GeoDistanceSlider`. What do you think? It should be about adding it here https://github.com/appbaseio/reactivesearch/blob/fe672a2009d5481fb09c195e69486f90f54a9c15/packages/maps/src/components/basic/GeoDistanceSlider.js#L85 as well as to `GeoDistanceDropdown`

Related: https://developers.google.com/maps/documentation/javascript/places-autocomplete

jankoritak:
Minimal implementation that closes #991.

### The problem

- At the moment both **GeoDistanceSlider** and **GeoDistanceDropdown** components only support control of a single [AutocompleteService.getPlacePredictions](https://developers.google.com/maps/documentation/javascript/reference/places-autocomplete-service#AutocompleteService.getPlacePredictions) request options property, namely `componentRestrictions: { country: restrictedCountries }`.
- The request options supports way more properties than `componentRestrictions`, that are current not accessible, viz. [the docs](https://developers.google.com/maps/documentation/javascript/reference/places-autocomplete-service#AutocompletionRequest)

### Proposed/Draft solution

- Both of the components now accept a property `serviceOptions` that contains config object for the call
- The solution is very minimal as I'd like to avoid breaking changes to the API. Will add Storybook examples and potentially stronger typings before further actions.

### DIFF
--- a/packages/maps/src/components/basic/GeoDistanceDropdown.js
+++ b/packages/maps/src/components/basic/GeoDistanceDropdown.js
@@ -315,6 +315,7 @@ class GeoDistanceDropdown extends Component {
 				{
 					input: value,
 					componentRestrictions: { country: restrictedCountries },
+					...this.props.serviceOptions,
 				},
 				(res) => {
 					const suggestionsList
@@ -498,6 +499,7 @@ GeoDistanceDropdown.propTypes = {
 	onValueChange: types.func,
 	placeholder: types.string,
 	react: types.react,
+	serviceOptions: types.props,
 	showFilter: types.bool,
 	showIcon: types.bool,
 	style: types.style,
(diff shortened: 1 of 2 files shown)


lakhansamani (approved):
Looks good to me 💯 
@jankoritak  Thank you for adding PR for this 👍

lakhansamani:
@jankoritak sorry it took is bit long, since we were coming up with some breaking changes. Above changes are released with `3.0.0-alpha.9`

PR #1003, opened 2019-05-29
feat(web): add support for unrated results in RatingsFilter

ShahAnuj2610:
https://github.com/appbaseio/reactivesearch/issues/267

### DIFF
--- a/packages/web/src/components/range/RatingsFilter.js
+++ b/packages/web/src/components/range/RatingsFilter.js
@@ -55,12 +55,7 @@ class RatingsFilter extends Component {
 		const hasMounted = false;
 
 		if (currentValue) {
-			this.setValue({
-				value: currentValue,
-				props,
-				hasMounted,
-				includeUnrated: currentValue[2],
-			});
+			this.setValue({ value: currentValue, props, hasMounted });
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jyash97:
Hey, @ShahAnuj2610 Thanks for the PR.

Have you tested this by keeping `URLParams` as `true` ?

ShahAnuj2610:
> Hey, @ShahAnuj2610 Thanks for the PR.
> 
> Have you tested this by keeping `URLParams` as `true` ?

@jyash97  fixed!

jyash97 (commented):
@ShahAnuj2610 

The feature works well but there are some UI changes and testing needed while implementing this. Here are a few things that I noticed:

1. When I don't have `includeUnrated`, I see an extra comma in the tab. Check this: 
image

2. When we don't have `includeUnrated` the URL params just says `true` or `null` so either we should show `includedUnrated` or doesn't show. cc @siddharthlatest Thoughts?

3. There is a proptype error coming on selectedValue

4. We will also need to document these changes over [here](https://opensource.appbase.io/reactive-manual/range-components/ratingsfilter.html)

siddharthlatest:
For 2., we don't need a URL query string for includeUnrated.

-Siddharth

jyash97 (approved):
@ShahAnuj2610  

These changes look good. Great Work! 💯 

Thanks for Contribution 🥂

PR #1010, opened 2019-06-11
Pass up index prop to onData in ReactiveList

nathan-charles:
This seems useful if you need to render ad's in between some rows


bietkul (approved):

siddharthlatest:
@bietkul Is this good to be merged?

bietkul (approved):

PR #1030, opened 2019-06-29
Update to new issue templates

siddharthlatest:
* Use new issue templates instead of the legacy template.
* Separate templates for bug and feature request.
* Better clarify the expectations when a user is either filing a bug report or a feature request.


siddharthlatest:
@bietkul Ideally, we should also remove the legacy template along with adding these.

bietkul (approved):

PR #1044, opened 2019-07-10
feat: Range components accessibility

jyash97:


packages/web/src/components/range/MultiRange.js
```diff
@@ -240,16 +240,23 @@ class MultiRange extends Component {
...
+					role="listbox"
+					aria-label={`${this.props.componentId}-items`}
+				>
 					{this.props.data.map(item => (
 						<li
 							key={item.label}
 							className={`${this.state.selectedValues[item.label] ? 'active' : ''}`}
+							role="option"
+							aria-checked={!!this.state.currentValue[item.label]}
+							aria-selected={!!this.state.currentValue[item.label]}
```
bietkul: Can you keep a variable to store `!!this.state.currentValue[item.label]`, and use it further.

bietkul (approved):

packages/web/src/components/range/MultiRange.js
bietkul: @jyash97 Can you update it?

packages/web/src/components/range/MultiRange.js
```diff
@@ -240,29 +240,43 @@ class MultiRange extends Component {
...
+					aria-label={`${this.props.componentId}-items`}
+				>
+					{this.props.data.map((item) => {
+						const isChecked = !!this.state.selectedValues[item.label];
+						return (
+							<li
+								key={item.label}
+								className={`${
+									this.state.selectedValues[item.label] ? 'active' : ''
+								}`}
```
bietkul, line 255: @jyash97 What's the point of adding a variable if you're not using it.

bietkul (changes requested):

packages/web/src/components/range/MultiRange.js
jyash97: Missed this line will fix it

PR #1082, opened 2019-08-07
fix(site): spacing and image size

jyash97:


siddharthlatest (changes requested):
- [x] We need to check on the home screen hero image's alignment with certain resolutions.

bietkul (approved):

PR #1090, opened 2019-08-14
feat(map): support exhaustive service options prop for geodistance

lakhansamani:
Support other library options for `next` version of `GeoDistanceSlider` and `GeoDistanceDropDown`

Ref: #996

### DIFF
--- a/packages/maps/src/components/basic/GeoDistanceSlider.js
+++ b/packages/maps/src/components/basic/GeoDistanceSlider.js
@@ -584,7 +584,7 @@ GeoDistanceSlider.propTypes = {
 	value: types.selectedValue,
 	unit: types.string,
 	URLParams: types.bool,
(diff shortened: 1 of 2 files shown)


packages/maps/src/components/basic/GeoDistanceSlider.js
```diff
@@ -583,6 +584,7 @@ GeoDistanceSlider.propTypes = {
 	value: types.selectedValue,
 	unit: types.string,
 	URLParams: types.bool,
+	...this.props.serviceOptions,
```
bietkul: @lakhansamani Does it not throw any error?

bietkul (changes requested):
Also, you have not added this prop in .ts files.

bietkul (approved):

PR #1095, opened 2019-08-16
Fix `setSearchParams` type definition

prayerslayer:
Hi there,

I just stumbled across this, it seems to me that the signature of `setSearchParams` in `ReactiveBase` is wrong. Probably a copy/paste error from the `get` equivalent. It should take the new URL and return nothing, not the other way around.

Hope I got it right!


bietkul (approved):

bietkul:
@prayerslayer Thanks for the fix.

PR #1102, opened 2019-08-19
fix(web): avoid re rendering of result item, pagination & poweredBy Image

jyash97:
ReactiveList-Performance

### DIFF
--- a/packages/web/src/components/result/addons/Pagination.js
+++ b/packages/web/src/components/result/addons/Pagination.js
@@ -23,7 +23,7 @@ const buildPaginationDOM = (props, position) => {
 	let start
 		= position === 'start'
 			? getStartPage(pages, currentPage, showEndPage)
-			: Math.ceil(totalPages - (pages - 1) / 2) + 1;
+			: Math.ceil(totalPages - ((pages - 1) / 2)) + 1;
 	const paginationButtons = [];
 	if (start <= totalPages) {
 		let totalPagesToShow = pages < totalPages ? start + (pages - 1) : totalPages + 1;
@@ -33,8 +33,8 @@ const buildPaginationDOM = (props, position) => {
 					? start + (Math.ceil(pages / 2) - (pages % 2))
 					: totalPages + 1;
 		}
-		if (currentPage > totalPages - pages + 2) {
-			start = totalPages - pages + 2;
+		if (currentPage > (totalPages - pages) + 2) {
+			start = (totalPages - pages) + 2;
 		}
 		for (let i = start; i < totalPagesToShow; i += 1) {
 			const primary = currentPage === i - 1;
@@ -145,7 +145,7 @@ class Pagination extends React.PureComponent {
 				{showEndPage && currentPage >= Math.floor(pages / 2) + !!(pages % 2) ? (
 					<span>...</span>
 				) : null}
-				{currentPage <= totalPages - pages + 2 && buildPaginationDOM(this.props, 'start')}
+				{currentPage <= (totalPages - pages) + 2 && buildPaginationDOM(this.props, 'start')}
 				{showEndPage && pages > 2 && currentPage <= totalPages - Math.ceil(pages * 0.75) ? (
 					<span>...</span>
 				) : null}
--- a/packages/web/src/components/result/addons/Pagination.js
+++ b/packages/web/src/components/result/addons/Pagination.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { getClassName, handleA11yAction } from '@appbaseio/reactivecore/lib/utils/helper';
+import { getClassName, handleA11yAction, isEqual } from '@appbaseio/reactivecore/lib/utils/helper';
 import types from '@appbaseio/reactivecore/lib/utils/types';
 
 import Button, { pagination } from '../../../styles/Button';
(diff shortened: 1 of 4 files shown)


packages/web/src/components/result/addons/Pagination.js
```diff
@@ -163,6 +164,12 @@ export default function Pagination(props) {
 	);
 }
 
+function areEqual(prevProps, nextProps) {
+	return isEqual(prevProps, nextProps);
+}
+
+export default React.memo(Pagination, areEqual);
```
bietkul: React.memo is not supported by React version < 16.6

bietkul (changes requested):

packages/web/src/components/result/addons/Pagination.js
jyash97: True. 
Should I change to `class` Component and use `PureComponent` ?

bietkul: @jyash97 Yes, you can try it out.
Btw what is the reason of extra re-rendering?

jyash97: The reason for re-rendering is due to redux store: 

- Promoted Results
- Loading
- Hits Update
- URLParams Update

We can't stop this re-rendering as this can cause some side effects.

jyash97:
Profiler data after applying `PureComponent` and `shouldComponentUpdate` changes:
image

jyash97:
@lakhansamani @bietkul  Is this good to go?

packages/web/src/components/result/addons/Pagination.js
```diff
@@ -65,104 +65,108 @@ const buildPaginationDOM = (props, position) => {
...
+					Next
+				</Button>
+			</div>
+		);
+	}
 }
 
+export default Pagination;
+
 Pagination.propTypes = {
```
bietkul, line 168: @jyash97 Can you put that block before export statement?

packages/web/src/components/result/addons/Results.js
```diff
@@ -0,0 +1,44 @@
...
+
+Results.propTypes = {
+	hasCustomRender: types.boolRequired,
+	innerClass: types.style,
+	renderItem: types.func,
+	base: types.number,
+	getComponent: types.func,
+	listClass: types.string,
+	triggerClickAnalytics: types.func,
+};
```
bietkul, line 44: @jyash97 `filteredResults` is not added in props validations, don't you have eslint running?
I'm not sure how we can ignore the eslint errors, we already set a pre-commit hook to avoid the linting errors.
We shouldn't bypass it.

packages/web/src/components/result/addons/Results.js
```diff
@@ -0,0 +1,44 @@
...
+import types from '@appbaseio/reactivecore/lib/utils/types';
+
+class Results extends React.Component {
+	shouldComponentUpdate(nextProps) {
+		/*
+			Use shouldComponentUpdate because PureComponent uses shallow comparison of objects.
+			If props contain complex data structures, it may produce false-negatives for
+			deeper differences.
+		*/
+		return !isEqual(nextProps, this.props);
```
bietkul: Would it not create performance issues, suppose you're rendering 50+ results in a page.
It'll be slow to compare props (results) with nested values.
I think PureComponent will be enough.

bietkul (changes requested):

packages/web/src/components/result/addons/Results.js
jyash97: I have added the comment for the same do check it. 

I used `PureComponent` component before but it was re-rendering the result.

packages/web/src/components/result/addons/Pagination.js
jyash97: Okay

packages/web/src/components/result/addons/Results.js
jyash97: Checking

packages/web/src/components/result/addons/Results.js
jyash97: Here is the screen of Profiler and console in both scenario for 100 Results:

#### With `shouldComponentUpdate`
image
image


#### With `PureComponent`
image
image

#### With comparing only `filteredResults` in `shouldComponentUpdate`
image
image

packages/web/src/components/result/addons/Pagination.js
```diff
@@ -33,8 +33,8 @@ const buildPaginationDOM = (props, position) => {
 					? start + (Math.ceil(pages / 2) - (pages % 2))
 					: totalPages + 1;
 		}
-		if (currentPage > (totalPages - pages) + 2) {
-			start = (totalPages - pages) + 2;
+		if (currentPage > totalPages - pages + 2) {
+			start = totalPages - pages + 2;
```
bietkul: @jyash97 Is your eslint with prettier off, this line has the linting error?
Why there are a lot of unnecessary changes?
Can you update it?

bietkul (changes requested):

packages/web/src/components/result/addons/Pagination.js
jyash97: There are eslint errors on this file and were there from before I didn't fix them because it would have created more noise in the PR we can do a follow-up PR to fix these errors. They are mainly because of `no-mix-operators`

bietkul (approved):

Issue #1057, opened 2019-07-16
Allow customizing the Dropdown label similarly to the items


therealtbs:
**Affected Projects**
React

**Is your feature request related to a problem? Please describe.**
I'm translating the items that are returned by Elasticsearch before displaying them to the user. For the actual dropdown list the `renderItem`-prop works great, however once the item is selected it displays the untranslated value from Elasticsearch in the dropdown label.

**Describe the solution you'd like**
Ideally there would be some sort of `renderLabel`-prop similar to the `renderItem`-prop to customize the display of the dropdown label without affecting the actual value that is used behind the scenes.

**Describe alternatives you've considered**
I have briefly considered using the `render`-prop to render the entire component myself, however that for one seems like total overkill for what I want to do and also would not have worked because the `render`-prop only influences what's rendered inside the opened dropdown.

**Additional context**
See this screenshot.
![image](https://user-images.githubusercontent.com/944183/61293474-09c01780-a7d4-11e9-9e1f-12fe6d3fb4c7.png)

I would like to have the '10' also be a nice translated value.

therealtbs:
Fixed in version 3.0.0

PR #1103, opened 2019-08-19
feat(web): renderLabel prop in Single- and MultiDropdownList (#1057)

therealtbs:
I took a stab at implementing my proposed solution for #1057.  

This PR adds a `renderLabel`-prop to render the label of the dropdown similarly to the `renderItems`-prop. Allowing for better UX by being able to show translated values as opposed to just the 'raw' items from the database.

I wanted this function to be able to return React elements as well as strings which posed the interesting challenge of whether or not the `title`-attribute should be influenced by the `renderLabel`-prop. The solution I settled on was to use the return value of the function, if that return value is a string, otherwise fall back to the normal behavior.

There were some linting errors in `Dropdown.js` complaining of missing prop-types, which I had to add in order to commit. I hope I guessed those two types correctly.

Let me know if you have any comments or changes.

### DIFF
--- a/packages/web/src/components/list/MultiDropdownList.d.ts
+++ b/packages/web/src/components/list/MultiDropdownList.d.ts
@@ -19,6 +19,7 @@ export interface MultiDropdownList extends CommonProps {
 	react?: types.react;
 	render?: (...args: any[]) => any;
 	renderItem?: (...args: any[]) => any;
+	renderLabel?: (...args: any[]) => any;
 	renderError?: types.title;
 	transformData?: (...args: any[]) => any;
 	selectAllLabel?: string;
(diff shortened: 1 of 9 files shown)


packages/web/src/components/list/MultiDropdownList.d.ts
```diff
@@ -19,6 +19,7 @@ export interface MultiDropdownList extends CommonProps {
 	react?: types.react;
 	render?: (...args: any[]) => any;
 	renderItem?: (...args: any[]) => any;
+	renderLabel?: (items: {[key: string]: boolean}, placeholder: string) => React.ReactElement;
```
bietkul: @therealtbs Any specific reason to add `placeholder` as the second parameter?
I think we can remove it.

bietkul: Please use the same signature.
`renderLabel?: (...args: any[]) => any;`

bietkul (approved):

bietkul (changes requested):
PR looks good.
Just left the few comments.
Can you please add `renderLabel` to other components which use `Dropdown`?
For example: MultiDropdownRange and SingleDropdownRange.

therealtbs:
I've added the prop to both of the Range components and changed the signature to remove the placeholder.

bietkul (approved):

Issue #1096, opened 2019-08-16
Unable to Remove Query / DynamicRangeSlider enforces value once active


janus-reith:
**Affected Projects**
React

**Library Version:** x.y.z
3.0.0-rc.25

**Describe the bug**
As soon as it rendered, the DynamicRangeSlider will enforce the range it displays. Items that don't have this field will not be matched anymore at all.  In my case, that would be an optional memory filter for devices, while not all devices do have any memory.

The SelectedFilters component however doesn't show any hint yet, that a range query is applied.

Next I put rendering of the component under a boolean condition that I stored in state and added a toogle for that. Now as soon as I enable the toggle and the DynamicRange becomes active, the restriction is applied. When I toggle it off, the query stays restricted to that range. I first have to change another value in order to update the results without the range.

**To Reproduce**
I dont have quick sandbox example, might provide one later.

**Expected behavior**
1. If a Range is queried, always show it in `SelectedFilters`.
2. When the component is removed, remove the range from the query.

3. Idea to be more intuitive:  Create a default disabled state for the DynamicRangeSlider, where the knobs are positioned at the min/max like they are now, but greyed out. Enable both when one of them has been dragged or clicked the first time, show some button that allows unselecting it again.

siddharthlatest:
@janus-reith A codesandbox (when you can) would help a lot. We're looking into the set of issues related to DynamicRangeSlider next and can check on this.

janus-reith:
The reason that I didn't provide a sandbox yet is that I would first need to have a public accessible dataset available which has a field present on only some items which can also be queried by range.

janus-reith:
I think this is related: https://github.com/appbaseio/reactivesearch/issues/1059

Ah, just noted that @ShahAnuj2610 already referenced the PR that fixes this, great!

Still, I'm wondering:

- How would I go about this if I manually want to remove the query somehow that the DynamicRangeSlider created?(Or any other query)  I didn't find a way for this yet.
I also tried to use the `connect` handle and remove, that works for setQuery but did not work properly for removing, even if the relevant component unmounted.

- Should there maybe be a visual distinction between a non selected range (greyed out slider handles) and actually selected minimum/maximum?

ShahAnuj2610:
> I think this is related: #1059
> 
> Ah, just noted that @ShahAnuj2610 already referenced the PR that fixes this, great!
> 
> Still, I'm wondering:
> 
[...]

PR #1108, opened 2019-08-20
feat(web): include `includeNullValues` prop to handle sparse data

### LINKED ISSUE #1059: Do not apply pristine range filters
**Affected Projects**
React

**Library Version:** x.y.z
3.0.0-rc.20

**Describe the bug**
When filtering sparse data (i.e., when some of the attributes can be present, but may not be), you want filter components to only be applied when the user explicitly sets them.

For example: I work on a search app that filters by some date field, but that date field is only present in 40% of the documents. When I render the `RangeSlider` component, only items are shown that contain some value for the `dataField` attribute. This means that only documents are shown that have that specific date field.

This happens, because the `RangeSlider` applies a filter to the query with the `range` `start` and `end` values.

Ideally, this filter is only applied after the user interacts with the component. This could be fixed by adding a `pristine` state to `RangeSlider` and others, and only applies the filter when its value is false.

Perhaps this could be set as an optional prop, e.g. `ignoreUntouched`.

ShahAnuj2610:
fixes #1059 #1096 
- feat - add support in all range components for the prop `includeNullValues`

### DIFF
--- a/packages/web/src/components/range/RangeInput.js
+++ b/packages/web/src/components/range/RangeInput.js
@@ -185,6 +185,7 @@ RangeInput.propTypes = {
 	style: types.style,
 	themePreset: types.themePreset,
 	componentId: types.stringRequired,
+	includeNullValues: types.bool,
 };
 
 RangeInput.defaultProps = {
@@ -193,6 +194,7 @@ RangeInput.defaultProps = {
 		end: 10,
 	},
 	stepValue: 1,
+	includeNullValues: false,
 };
 
 const mapStateToProps = state => ({
--- a/packages/web/src/components/range/RatingsFilter.js
+++ b/packages/web/src/components/range/RatingsFilter.js
@@ -121,12 +121,12 @@ class RatingsFilter extends Component {
 		return value ? [value.start, value.end] : null;
 	};
 
-	static defaultQuery = (value, props, includeNullValues = false) => {
+	static defaultQuery = (value, props, includeUnrated = false) => {
 		let query = null;
 		if (value) {
 			query = getRangeQueryWithNullValues(value, {
 				dataField: props.dataField,
-				includeNullValues: props.includeNullValues || includeNullValues,
+				includeNullValues: props.includeNullValues || includeUnrated,
 			});
 		}
 
@@ -174,9 +174,9 @@ class RatingsFilter extends Component {
 		checkValueChange(props.componentId, value, props.beforeValueChange, performUpdate);
 	};
 
-	updateQuery = (value, props, includeNullValues) => {
+	updateQuery = (value, props, includeUnrated) => {
 		const { customQuery } = props;
-		let query = RatingsFilter.defaultQuery(value, props, includeNullValues);
+		let query = RatingsFilter.defaultQuery(value, props, includeUnrated);
 		let customQueryOptions;
 		if (customQuery) {
 			({ query } = customQuery(value, props) || {});
--- a/packages/web/src/components/range/RatingsFilter.js
+++ b/packages/web/src/components/range/RatingsFilter.js
@@ -28,7 +28,7 @@ import Title from '../../styles/Title';
 import Container from '../../styles/Container';
 import StarRating from './addons/StarRating';
 import { ratingsList } from '../../styles/ratingsList';
-import { connect, getValidPropsKeys } from '../../utils';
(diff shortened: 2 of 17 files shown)


packages/web/src/components/range/RangeInput.js
```diff
@@ -185,6 +185,7 @@ RangeInput.propTypes = {
 	style: types.style,
 	themePreset: types.themePreset,
 	componentId: types.stringRequired,
+	includeNullValues: types.bool,
```
bietkul, line 188: @ShahAnuj2610 Are we missing the query generation here?

packages/web/src/components/range/RatingsFilter.js
```diff
@@ -134,13 +139,13 @@ class RatingsFilter extends Component {
 				},
 			};
 			if (includeNullValues) {
```
bietkul: I think we're already checking it in the util function.

bietkul (changes requested):
Typings are missing for `includeNullValues` in `.ts` files for all the components.

packages/web/src/components/range/RangeInput.js
ShahAnuj2610: can you elaborate?

bietkul: Ok, I think it is not needed.

packages/web/src/components/range/RatingsFilter.js
```diff
@@ -129,23 +124,10 @@ class RatingsFilter extends Component {
 	static defaultQuery = (value, props, includeNullValues = false) => {
```
bietkul: @ShahAnuj2610 Why `includeNullValues` is the third argument?
We can access it from `props`.

bietkul (changes requested):

packages/web/src/components/range/RatingsFilter.js
ShahAnuj2610: no, the third argument comes from the `includeUnrated` in the data object
`
			const handleUpdates = () => {
				this.updateQuery(value, props, includeUnrated);
				this.locked = false;
				if (props.onValueChange) props.onValueChange(value);
			}
`

ShahAnuj2610: https://github.com/appbaseio/reactivesearch/pull/1003

bietkul: But why you named it `includeNullValues`?

bietkul (changes requested):

ShahAnuj2610:
@bietkul can rename `includeNullValues` param in func to `includeUnrated`

packages/web/src/components/range/RatingsFilter.js
```diff
@@ -121,12 +121,12 @@ class RatingsFilter extends Component {
...
 	};
 
-	static defaultQuery = (value, props, includeNullValues = false) => {
+	static defaultQuery = (value, props, includeUnrated = false) => {
 		let query = null;
 		if (value) {
 			query = getRangeQueryWithNullValues(value, {
 				dataField: props.dataField,
-				includeNullValues: props.includeNullValues || includeNullValues,
+				includeNullValues: props.includeNullValues || includeUnrated,
```
bietkul, line 129: @ShahAnuj2610 I'm just curious to know that how these(`includeNullValues` & `includeUnrated`) props are related to each other?

bietkul (changes requested):

packages/web/src/components/range/RatingsFilter.js
ShahAnuj2610: the query executed for both these parameters are same, only `includeUnrated` is applied at the item level, whereas `includeNullValues` is applied for every item in the data
`
<RatingsFilter
					componentId="RatingsSensor"
					dataField="average_rating_rounded"
					title="RatingsFilter"
					data={[
						{ start: 4, end: 5, label: '4 stars and up', includeUnrated: true },
						{ start: 3, end: 5, label: '3 stars and up' },
						{ start: 2, end: 5, label: '2 stars and up' },
						{ start: 1, end: 5, label: '> 1 stars' },
					]}
				/>
`
In the above if `includeNullValues=false`, then null results will only be included for the first data item

bietkul (approved):

PR #1111, opened 2019-08-22
feat(web): handle controlled behavior of component

### LINKED ISSUE #1091: onChange handler and value behave weird and not consistent
**Affected Projects**
React

**Library Version:** x.y.z
v3.0.0-rc.25

**Describe the bug**
`onChange` handlers only become active when the component has a `value` set.
While i don't understand why this limitation is in place, the real issue for me is that `onChange` does not return a value that I can use for `value`.
E.g. MultiList returns a single string upon change, expects an array and internally uses an object to keep the current value in `state`.
For the RangeInput I had similar issues.

**To Reproduce**
Take `MultiList` and add `value` from state, and pass a function to `onChange` that sets the value to state. Actually, you wouldn't even need to set it and log it to console instead.

**Expected behavior**
I can easily store the data from an `onChange` in my global store and pass it back to the component using `value`

ShahAnuj2610:
fixes #1091

### DIFF
--- a/packages/web/src/components/list/MultiDataList.js
+++ b/packages/web/src/components/list/MultiDataList.js
@@ -112,7 +112,7 @@ class MultiDataList extends Component {
 		}
 
 		if (this.props.value !== prevProps.value) {
-			this.setValue(this.props.value, true);
+			this.setValue(this.props.value || [], true);
 		} else if (
 			!isEqual(selectedValue, this.props.selectedValue)
 			&& !isEqual(this.props.selectedValue, prevProps.selectedValue)
@@ -385,8 +385,11 @@ class MultiDataList extends Component {
 		if (value === undefined) {
 			this.setValue(currentValue);
 		} else if (onChange) {
-			this.setValue(currentValue);
-			onChange(Object.keys(this.state.currentValue));
+			const newValue = Object.assign([], this.props.value);
+			const currentValueIndex = newValue.indexOf(currentValue);
+			if (currentValueIndex > -1) newValue.splice(currentValueIndex, 1);
+			else newValue.push(currentValue);
+			onChange(newValue);
 		}
 	};
 
--- a/packages/web/src/components/list/MultiDataList.js
+++ b/packages/web/src/components/list/MultiDataList.js
@@ -385,7 +385,8 @@ class MultiDataList extends Component {
 		if (value === undefined) {
 			this.setValue(currentValue);
 		} else if (onChange) {
-			onChange(currentValue);
+			this.setValue(currentValue);
+			onChange(Object.keys(this.state.currentValue));
 		}
 	};
 
--- a/packages/web/src/components/list/MultiDataList.js
+++ b/packages/web/src/components/list/MultiDataList.js
@@ -29,7 +29,14 @@ import Title from '../../styles/Title';
 import Input from '../../styles/Input';
 import Container from '../../styles/Container';
 import { UL, Checkbox } from '../../styles/FormControlList';
-import { connect, getComponent, hasCustomRenderer, isEvent, getValidPropsKeys } from '../../utils';
+import {
+	connect,
+	getComponent,
+	hasCustomRenderer,
+	isEvent,
+	getValidPropsKeys,
+	parseValueArray,
+} from '../../utils';
 import { getAggsQuery } from './utils';
 
 class MultiDataList extends Component {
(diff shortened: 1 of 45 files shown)


bietkul:
@ShahAnuj2610 You are missing range components.

packages/web/src/components/list/MultiDataList.js
```diff
@@ -385,7 +385,8 @@ class MultiDataList extends Component {
 		if (value === undefined) {
 			this.setValue(currentValue);
 		} else if (onChange) {
-			onChange(currentValue);
+			this.setValue(currentValue);
```
bietkul: @ShahAnuj2610 Why did you add `setValue` here?

packages/web/src/components/list/MultiDropdownList.js
```diff
@@ -417,7 +415,8 @@ class MultiDropdownList extends Component {
 		if (value === undefined) {
 			this.setValue(currentValue);
 		} else if (onChange) {
-			onChange(currentValue);
+			this.setValue(currentValue);
+			onChange(Object.keys(this.state.currentValue));
```
bietkul: The same issue is here.

packages/web/src/components/list/MultiList.js
```diff
@@ -457,7 +457,8 @@ class MultiList extends Component {
 		if (value === undefined) {
 			this.setValue(currentValue);
 		} else if (onChange) {
-			onChange(currentValue);
+			this.setValue(currentValue);
+			onChange(Object.keys(this.state.currentValue));
```
bietkul: The same issue is here.

packages/web/src/components/range/MultiDropdownRange.js
```diff
@@ -220,9 +238,12 @@ class MultiDropdownRange extends Component {
 		const { value, onChange } = this.props;
 
 		if (value === undefined) {
-			this.selectItem(items);
+			this.selectItem({ item: items });
 		} else if (onChange) {
-			onChange(items);
+			this.selectItem({
```
bietkul: We don't need to create any side effect when controlled behavior is being used, Just pass the value to the `onChange`.

bietkul (changes requested):
I think the current implementation is not correct.
When the user defines the `value` with `onChange` then it is the user's responsibility to update the value prop accordingly.
It depends on the user whether he wants to update value or not.
Currently, you're just supplying the value to `onChange` after triggering the query i.e using `setValue`.

ShahAnuj2610:
> I think the current implementation is not correct.
> When the user defines the `value` with `onChange` then it is the user's responsibility to update the value prop accordingly.
> It depends on the user whether he wants to update value or not.
> Currently, you're just supplying the value to `onChange` after triggering the query i.e using `setValue`.

@bietkul refactoring done

jyash97 (changes requested):
@ShahAnuj2610  Can we avoid using Controlled Component in Examples as they are directly linked to docs via Codesandbox Embed links.

ShahAnuj2610:
> @ShahAnuj2610 Can we avoid using Controlled Component in Examples as they are directly linked to docs via Codesandbox Embed links.

Since there is a behavioral change in the `onChange` method, shouldn't we include an example, just so the user is aware?

bietkul:
I agree we should create different examples for controlled components.
For example, we already have `DataSearchControlled` example for `DataSearch`.

ShahAnuj2610:
> I agree we should create different examples for controlled components.
> For example, we already have `DataSearchControlled` example for `DataSearch`.

examples added for controlled components

packages/web/src/components/list/MultiDataList.js
```diff
@@ -385,7 +385,11 @@ class MultiDataList extends Component {
 		if (value === undefined) {
 			this.setValue(currentValue);
 		} else if (onChange) {
-			onChange(currentValue);
+			const newValue = Object.assign([], this.props.value);
+			const currentValueIndex = newValue.indexOf(currentValue);
+			if (currentValueIndex > -1) newValue.splice(currentValueIndex, 1);
+			else newValue.push(currentValue);
```
bietkul: Put this whole logic in some util function, please avoid code duplication.

bietkul (changes requested):

bietkul (approved):

PR #1124, opened 2019-08-28
fix(web): force ReactiveList to re-render if customRender changes

therealtbs:
When using ReactiveList with a render function, the component does not update if the render function changes. (this also applies to `renderItem`)

This is problematic because the render function could depend on props from outside. When those props change, the update is swallowed by the Results component, since in it's eyes nothing changed.

I've added the relevant props to the `shouldComponentUpdate` function of the Results component.

Maybe there should be a note added to the docs about using `useCallback`/`useMemo` together with these functions in order to better control their update behavior.

### DIFF
--- a/packages/web/src/components/result/addons/Results.js
+++ b/packages/web/src/components/result/addons/Results.js
@@ -10,7 +10,12 @@ class Results extends React.Component {
 			deeper differences.
 		*/
 
-		return !isEqual(nextProps.filteredResults, this.props.filteredResults);
+		return !(
+			isEqual(nextProps.filteredResults, this.props.filteredResults)
+			&& isEqual(nextProps.hasCustomRender, this.props.hasCustomRender)
+			&& isEqual(nextProps.getComponent, this.props.getComponent)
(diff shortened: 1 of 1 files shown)


jyash97 (commented):
Hey @therealtbs Can you please specify an example of when the re render was not triggered?

therealtbs:
@jyash97 I've made a CodeSandbox which should highlight his issue nicely. https://codesandbox.io/s/vigilant-hamilton-g6wqo

It's the classic button counter example, but as you can see the display inside the ReactiveList only updates when the list of results changes.

bietkul:
@therealtbs Let's completely remove the use of `shouldComponentUpdate` for now, the current implementation can break the existing UI after an upgrade.
@jyash97 We have to find some other ways to deal with extra re-rendering.

therealtbs:
@bietkul I've removed it from the `Results` component. AFAICT the `Pagination` and `PoweredBy` components work correctly, so there is no need to completely revert #1102.

bietkul (approved):

PR #1125, opened 2019-08-29
fix(web): focus input after mic has stopped listening

jyash97:
Currently, after we get a search term  through speech recognition the input loses the focus

#### Category Search
CS
___
#### DataSearch
DS

### DIFF
--- a/packages/web/src/components/search/DataSearch.js
+++ b/packages/web/src/components/search/DataSearch.js
@@ -586,7 +586,11 @@ class DataSearch extends Component {
 			&& results[0][0].transcript.trim()
 		) {
 			this.isPending = false;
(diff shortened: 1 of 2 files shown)


packages/web/src/components/search/DataSearch.js
```diff
@@ -586,7 +586,11 @@ class DataSearch extends Component {
 			&& results[0][0].transcript.trim()
 		) {
 			this.isPending = false;
+			this._inputRef.focus();
```
bietkul: Just do it when `autosuggest` is `true`

bietkul (changes requested):

bietkul (approved):

PR #1134, opened 2019-09-03
fix(maps): reduce re-renders because of markers

lakhansamani:
* Improve google map performance by reducing re-renders because of markers position and state
* use componentDidUpdate instead of `componentWillUpdate`

Dependent on following PR
https://github.com/appbaseio/reactivecore/pull/45/files

### DIFF
--- a/packages/maps/src/components/result/GoogleMapMarker.js
+++ b/packages/maps/src/components/result/GoogleMapMarker.js
@@ -1,0 +1,223 @@
+import React from 'react';
+import { MarkerWithLabel } from 'react-google-maps/lib/components/addons/MarkerWithLabel';
+import { InfoWindow, Marker } from 'react-google-maps';
+
+import types from '@appbaseio/reactivecore/lib/utils/types';
+import { setMarkerOnTop, setOpenMarkers } from '@appbaseio/reactivecore/lib/actions';
+import { connect } from '@appbaseio/reactivesearch/lib/utils';
+
+import { MapPin, MapPinArrow, mapPinWrapper } from './addons/styles/MapPin';
+
+class GoogleMapMarker extends React.Component {
+	shouldComponentUpdate(nextProps) {
+		if (
+			nextProps.markerOnTop === this.props.marker._id
+			|| (this.props.marker._id === this.props.markerOnTop && nextProps.markerOnTop === null)
+		) {
+			return true;
+		}
+
+		if (
+			this.props.openMarkers[this.props.marker._id]
+			!== nextProps.openMarkers[this.props.marker._id]
+		) {
+			return true;
+		}
+
+		return false;
+	}
+
+	openMarker = () => {
+		const {
+			setOpenMarkers: handleOpenMarkers,
+			openMarkers,
+			marker,
+			autoClosePopover,
+			handlePreserveCenter,
+		} = this.props;
+		const id = marker._id;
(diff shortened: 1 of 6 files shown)


packages/maps/src/components/result/ReactiveGoogleMap.js
```diff
@@ -48,53 +47,25 @@ class ReactiveGoogleMap extends Component {
...
-			this.handleStyleChange(nextProps.defaultMapStyle);
+	componentDidUpdate(prevProps) {
+		if (this.props.defaultMapStyle !== prevProps.defaultMapStyle) {
+			this.handleStyleChange(this.props.defaultMapStyle);
 		}
 	}
 
 	handleStyleChange = (newStyle) => {
-		this.setState({
+		this.setState({});
```
jyash97: I think you missed to clean this line or is it intentional ?

packages/maps/src/components/result/GoogleMapMarkers.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import types from '@appbaseio/reactivecore/lib/utils/types';
+
+import GoogleMapMarker from './GoogleMapMarker';
+
+const GoogleMapMarkers = ({ showMarkers, resultsToRender, ...rest }) => (
+	<div>
+		{showMarkers
```
jyash97: How about checking this condition in the parent component?  In this way, we can stop the extra div to render

lakhansamani: Sure thing 👍

packages/maps/src/components/result/GoogleMapMarker.js
```diff
@@ -0,0 +1,223 @@
...
+import { InfoWindow, Marker } from 'react-google-maps';
+
+import types from '@appbaseio/reactivecore/lib/utils/types';
+import { setMarkerOnTop, setOpenMarkers } from '@appbaseio/reactivecore/lib/actions';
+import { connect } from '@appbaseio/reactivesearch/lib/utils';
+
+import { MapPin, MapPinArrow, mapPinWrapper } from './addons/styles/MapPin';
+
+class GoogleMapMarker extends React.Component {
+	shouldComponentUpdate(nextProps) {
```
bietkul, line 12: We're using `shouldComponentUpdate` but it may have UI issues when someone uses the `state` in the `renderData`.
Can we add one more prop by which we can bypass this behavior?
We're using `strict` prop in `StateProvider`.

packages/maps/src/components/result/ReactiveMap.js
```diff
@@ -245,89 +245,80 @@ class ReactiveMap extends Component {
...
 		}
 
 		if (
-			isEqual(this.props.hits, nextProps.hits)
-			&& isEqual(this.props.streamHits, nextProps.streamHits)
+			isEqual(this.props.hits, prevProps.hits)
+			&& isEqual(this.props.streamHits, prevProps.streamHits)
 		) {
 			return false;
 		}
```
bietkul: The same thing is here.

bietkul (changes requested):

packages/maps/src/components/result/GoogleMapMarker.js
lakhansamani: @bietkul what could be the issues, when someone uses `state` in `renderData`?

PR #1145, opened 2019-09-10
fix(web): trigger renderNoResults when no items returned from msearch

### LINKED ISSUE #1137: MultiDropDownList not triggering renderNoResults
**Affected Projects**
ReactJS

**Library Version:** x.y.z
3.0.0

**Describe the bug**
MultiDropDownList not triggering renderNoResults when there isn't any results

**To Reproduce**
` <MultiDropdownList
                    componentId="Test"
                    dataField="data.test.keyword"
                    size={100}
                    sortBy="asc"
                    showCount
                    placeholder="Selec"
                    showFilter
                    loader="Loading ..."
                    showSearch
                    innerClass={{
                      select: 'search-select',
                      list: 'search-list',
                    }}
                    renderNoResults={() => <p>No Results Found!</p>}
                  />`

**Expected behavior**
To see No Results Found! text instead of empty div

jyash97:
Closes #1137

### DIFF
--- a/packages/web/src/components/list/MultiDataList.js
+++ b/packages/web/src/components/list/MultiDataList.js
@@ -427,7 +427,7 @@ class MultiDataList extends Component {
 		const { options } = this.state;
 
 		if (!this.hasCustomRenderer && options.length === 0) {
-			return null;
+			return this.props.renderNoResults ? this.props.renderNoResults() : null;
 		}
 
 		const listItems = this.listItems;
--- a/packages/web/src/components/list/SingleDropdownList.js
+++ b/packages/web/src/components/list/SingleDropdownList.js
@@ -23,11 +23,7 @@ import {
 import { componentTypes } from '@appbaseio/reactivecore/lib/utils/constants';
 import types from '@appbaseio/reactivecore/lib/utils/types';
 
-import {
-	getAggsQuery,
-	getCompositeAggsQuery,
-	updateInternalQuery,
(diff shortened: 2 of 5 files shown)


packages/web/src/components/list/SingleDropdownList.js
```diff
@@ -344,6 +340,10 @@ class SingleDropdownList extends Component {
 		}
 
 		if (!this.hasCustomRenderer && this.state.options.length === 0) {
+			if (this.props.renderNoResults && !this.props.isLoading) {
```
jyash97, line 343: `Note For Reviewer`: Added loading condition because on starting when the items are being loaded it would trigger  `renderNoResults` which is not a good UX so adding a condition to bypass this.

packages/web/src/components/list/MultiDataList.js
```diff
@@ -427,7 +427,7 @@ class MultiDataList extends Component {
 		const { options } = this.state;
 
 		if (!this.hasCustomRenderer && options.length === 0) {
-			return null;
+			return this.props.renderNoResults ? this.props.renderNoResults() : null;
 		}
 
 		const listItems = this.listItems;
```
bietkul, line 433: @jyash97 `renderNoResults` depends on the value of `this.items` not on the `options`.

bietkul (changes requested):

bietkul (approved):

PR #1154, opened 2019-09-12
chore(vue): refactor code for ToggleButton

jyash97:
- Replace `this.$props.item` to `this.item`
- To access all props we may need to pass `this.$props`

### DIFF
--- a/packages/vue/src/components/list/ToggleButton.jsx
+++ b/packages/vue/src/components/list/ToggleButton.jsx
@@ -34,9 +34,8 @@ const ToggleButton = {
 		renderItem: types.func,
 	},
 	data() {
-		const props = this.$props;
-		const value = this.selectedValue || props.value || props.defaultSelected || [];
-		const currentValue = ToggleButton.parseValue(value, props);
+		const value = this.selectedValue || this.value || this.defaultSelected || [];
+		const currentValue = ToggleButton.parseValue(value, this.$props);
 		this.__state = {
 			currentValue,
 		};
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/list/ToggleButton.jsx
```diff
@@ -34,9 +34,8 @@ const ToggleButton = {
 		renderItem: types.func,
 	},
 	data() {
-		const props = this.$props;
-		const value = this.selectedValue || props.value || props.defaultSelected || [];
-		const currentValue = ToggleButton.parseValue(value, props);
+		const value = this.selectedValue || this.value || this.defaultSelected || [];
+		const currentValue = ToggleButton.parseValue(value, this.$props);
```
bietkul, line 38: Here you're passing `this.$props` in `parseValue`.

packages/vue/src/components/list/ToggleButton.jsx
```diff
@@ -45,73 +44,72 @@ const ToggleButton = {
...
 		});
 	},
 	beforeDestroy() {
-		this.removeComponent(this.$props.componentId);
+		this.removeComponent(this.componentId);
 	},
 	watch: {
 		defaultSelected(newVal) {
-			this.setValue(ToggleButton.parseValue(newVal, this.$props));
+			this.setValue(ToggleButton.parseValue(newVal));
```
bietkul, line 67: But `this.$props` is missing here, why is there inconsistency?

bietkul (changes requested):
@jyash97  Can you do one thing, don't modify the `this.$props` passed in the methods because it might break the things which we'll be hard to debug in the future.
for example, don't change things like that `this.updateQuery(this.$data.currentValue, this.$props);` to just `this.updateQuery(this.currentValue);`, you removed the second param completely which made the method less customizable.
What if I want to set something forcefully in second param?
It is not possible now.
We're doing such things in other components.
Just change the way of accessing the properties for example if somewhere we're using `this.$data.currentValue` or `this.$props.currentValue` then change it to just `this.currentValue`.

jyash97:
After rethinking again I guess we should avoid this in order to not break things and the current set of things are working well and we will need `this.$props` in order to pass all props value. So lets close this PR.

PR #1159, opened 2019-09-13
fix(web): selectedFilters clearAll behavior

### LINKED ISSUE #1155: [Feature Request] On <SelectedFilters>, showClearAll should always display the button
**Affected Projects**
React

**Is your feature request related to a problem? Please describe.**
I would like to have a button which allows me to remove all filters. However I don't want the filters cloud with  it. So, my solution was to do this: 

`<SelectedFilters showClearAll clearAllLabel="Remove all filters" />`

and put `showFilter={false}` on each other component.

However, if you do that, you'll never get the clearAll button.

**Describe the solution you'd like**
`showClearAll` should display the clear all button anytime you've got a filter, even if showFilter is to false on it.

**Describe alternatives you've considered**
Make a special clearFilter component to define what you want to clear: a componentId or all filters like http://docs.searchkit.co/v2.0.0/components/navigation/reset.html

ShahAnuj2610:
fixes #1155

### DIFF
--- a/packages/web/src/components/basic/SelectedFilters.js
+++ b/packages/web/src/components/basic/SelectedFilters.js
@@ -89,11 +89,11 @@ class SelectedFilters extends Component {
 	hasFilters = () =>
 		Object.keys(this.props.selectedValues)
 			.filter(id => this.props.components.includes(id))
-			.reduce((boolAcc, component) => {
-				const { label, value } = this.props.selectedValues[component];
+			.some((component) => {
+				const { value } = this.props.selectedValues[component];
 				const isArray = Array.isArray(value);
-				return boolAcc && label && ((isArray && value.length) || (!isArray && value));
-			}, true);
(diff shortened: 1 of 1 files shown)


packages/web/src/components/basic/SelectedFilters.js
```diff
@@ -83,28 +85,38 @@ class SelectedFilters extends Component {
 			.filter(Boolean);
 	};
 
+	// determines whether any filter has been applied regardless of `showFilter=false`
+	hasFilters = () =>
```
jyash97, line 89: If this is to only determine if we have filters applied than I think we should avoid looping to all keys instead we can break the loop if a filter is found. 

Maybe you can use `Array.some` instead of `reduce` 

Thoughts @bietkul @lakhansamani

jyash97 (approved):
LGTM 👍 
@lakhansamani @bietkul  any feedback or suggestion ?

PR #1167, opened 2019-09-20
fix(vue): reset of range values

ShahAnuj2610:
fixes https://github.com/appbaseio/reactivesearch/pull/988#issuecomment-494286016


jyash97 (changes requested):
This doesn't seem to work. Try to add `MultiList` component in the example and react dependency of List Component to Slider and try selecting an item and moving slider values. You will see after deselecting all items the slider doesn't update its values.

bietkul:
@ShahAnuj2610 Any update here?

ShahAnuj2610:
> @ShahAnuj2610 Any update here?

it's handled here #1189

PR #1183, opened 2019-10-03
Improve CategorySearch in web

### LINKED ISSUE #814: More than 2 suggested categories in categorySearch?
**Issue Type:**

I think this falls into the category of enhancement 

**Platform:**

Web

**Description:**

Is there currently a way, or is it possible for you to add more than just two suggested categories in the categorySearch?

screen shot 2019-02-01 at 4 53 55 pm

kvin97:
fix #814

### DIFF
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -1090,7 +1090,6 @@ CategorySearch.propTypes = {
 	autosuggest: types.bool,
 	beforeValueChange: types.func,
 	categoryField: types.string,
-	categorySuggestionsListSize: types.number,
 	className: types.string,
 	clearIcon: types.children,
 	componentId: types.stringRequired,
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -723,15 +723,15 @@ class CategorySearch extends Component {
 					{this.renderCancelIcon()}
 				</InputIcon>
 			)}
-			{this.props.showVoiceSearch
-				&& (
-					<Mic
-						getInstance={this.props.getMicInstance}
-						render={this.props.renderMic}
-						iconPosition={this.props.iconPosition}
-						onResult={this.handleVoiceResults}
-						className={getClassName(this.props.innerClass, 'mic') || null}
-					/>)}
+			{this.props.showVoiceSearch && (
+				<Mic
+					getInstance={this.props.getMicInstance}
+					render={this.props.renderMic}
+					iconPosition={this.props.iconPosition}
+					onResult={this.handleVoiceResults}
+					className={getClassName(this.props.innerClass, 'mic') || null}
+				/>
+			)}
 			<InputIcon onClick={this.handleSearchIconClick} iconPosition={this.props.iconPosition}>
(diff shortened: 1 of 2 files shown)


packages/web/src/components/search/CategorySearch.js
```diff
@@ -858,32 +859,25 @@ class CategorySearch extends Component {
...
-				{
-					label: `${this.state.currentValue} in all categories`,
-					value: this.state.currentValue,
-					category: '*',
-					// no source object exists for category based suggestions
-					source: null,
-				},
-				{
-					label: `${this.state.currentValue} in ${filteredCategories[0].key}`,
+			const categorySuggestions = [];
```
jyash97, line 862: How about slicing the array of `filteredCategories` and then mapping over it. Something like this: 

```
const size = filteredCategories.length < categorySuggestionsListSize ?  filteredCategories.length  : categorySuggestionsListSize;
const categoriesToDisplay = filteredCatogories.slice(0, size).map((item) => {
   // push to array ( try using spread operator if possible )
});
```

jyash97: Also please try to test it locally rigorously.

jyash97:
Can you also follow the PR template? They are present for a reason so that we don't miss documentation or other things related to PR.

bietkul:
@jyash97 Can you review it?

packages/web/src/components/search/CategorySearch.js
```diff
@@ -993,8 +988,9 @@ class CategorySearch extends Component {
 										...rest,
 									})}
 								{!this.hasCustomRenderer
-									&& isOpen
-									&& finalSuggestionsList.length ? (
+								&& isOpen
+								&& finalSuggestionsList.length ? (
+										// eslint-disable-next-line react/jsx-indent
```
jyash97: I think we wont need to disable eslint. Any reason for this ?

jyash97 (changes requested):
LGTM 💯 

The only change we need is to revert eslint changes and the prop value will be a maximum of 10. We cant have more than that. You will need to change Prop type

kvin97:
I found the following error with the removal of disable eslint.
fix1
Once I fixed it, following lint errors popped up.
fix2
Both errors comes as `eslint(react/jsx-indent)`.
I found similar kind of issue here.
https://github.com/yannickcr/eslint-plugin-react/issues/1679

kvin97:
> LGTM 
> 
> The only change we need is to revert eslint changes and the prop value will be a maximum of 10. We cant have more than that. You will need to change Prop type

kvin97:
shall I use the following propType to specify the range, by importing PropTypes as `prop-types`.
`categorySuggestionsListSize: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),`

jyash97:
I think there is some eslint config issue on your side as I tried changing things on the same file and could not replicate. 

Also for `propTypes` you may need to define a custom prop type.

kvin97:
To remove the eslint issue I revert back to a previous commit and move the code changes again. 
For `propTypes` custom prop type was defined. 
And I tested it locally rigorously.

Issue #801, opened 2019-01-22
Vue pagination UX problem


mauxtin:
Current outcome when current page is 5 and there are more than 5 pages

Prev 1 2 3 4 **5** Next
![image](https://user-images.githubusercontent.com/40521145/51532811-cc448800-1e49-11e9-84d2-f9975d5d51fb.png)

Expected outcome when current page is 5 (Show that there are more than 5 pages)
Prev 1 ... 4 **5** 6 7 ... Next

Of course the Next button goes to sixth page as expected but it should somehow indicate to the user that there more pages to navigate after the fifth.

metagrover:
Thanks for bringing this up. This enhancement can also be applied to react library as well.

bietkul:
Something like that might be a better option

![image](https://i.imgur.com/jNVvxuf.png)

Reference: https://ant.design/components/pagination/

mauxtin:
Closing this since it has been implemented.

PR #1184, opened 2019-10-03
feat(vue): add showEndPage to ReactiveList

jyash97:


arcxyz:
This will close #801 

In terms of user experience I think this should be the default behaviour of the pagination instead of making it optional using the `showEndPage`. I know it's how it works on the React component but it would be great if the appbase team consider making it the default.

bietkul (approved):

PR #1189, opened 2019-10-07
DynamicRangeSlider fix

ShahAnuj2610:
- extending #988 
- fixes reset behaviour of DynamicRangeSlider


site/src/components/Home.js
```diff
@@ -65,7 +65,7 @@ class HomePage extends Component {
 		// To fetch reactive search github stars
 		fetch('https://api.github.com/repos/appbaseio/reactivesearch')
 			.then(res => res.json())
-			.then(res => {
+			.then((res) => {
```
bietkul, line 68: Why there are changes in this file?

bietkul (changes requested):
@ShahAnuj2610 I'm not sure, what is exactly this PR doing.
I can see the changes in `vue` & `react` folder which is confusing.
alt text

site/src/components/Home.js
ShahAnuj2610: I back merged `next` and tried to commit, and it was throwing `eslint` errors for the above files, so had to fix it

packages/vue/src/components/range/DynamicRangeSlider.jsx
```diff
@@ -1,12 +1,11 @@
 import VueTypes from 'vue-types';
+import VueSlider from 'vue-slider-component';
 import { Actions, helper } from '@appbaseio/reactivecore';
-import NoSSR from 'vue-no-ssr';
 import Container from '../../styles/Container';
```
bietkul, line 4: You removed `vue-no-ssr`, Please make sure it works well with the server-side rendering.
Just test the `with-nuxt` example.
If it works then remove this package from `RangeSlider` component & `package.json`.

bietkul (changes requested):

bietkul:
@ShahAnuj2610 It lgtm, just make sure that it works well with SSR.

bietkul:
Please also verify the issues commented by @mauxtin at [here](https://github.com/appbaseio/reactivesearch/pull/988#issuecomment-500521131).

ShahAnuj2610:
@bietkul SSR and URLParams issues have been fixed

packages/vue/examples/with-nuxt/nuxt.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
 	/*
 	 ** Plugins to load before mounting the App
 	 */
-	plugins: ['~/plugins/reactivesearch-vue'],
+	plugins: [{ src: '~/plugins/reactivesearch-vue', ssr: false }],
```
bietkul: It doesn't solve the issue, it means that you're using ReactiveSearch without SSR.
We have to support SSR.

bietkul (changes requested):

packages/vue/examples/with-nuxt/nuxt.config.js
ShahAnuj2610: https://github.com/NightCatSama/vue-slider-component/issues/114#issuecomment-353827607

bietkul: I already know that but it is not gonna work.
Try some other thing, if doesn't work then don't upgrade the `vue-slider-component`.

bietkul: @ShahAnuj2610 You can try it once https://nightcatsama.github.io/vue-slider-component/#/?hash=server-side-rendering-ssr

ShahAnuj2610:
revert `vue-slider-component` to `2.8.2` for `SSR` to work, here is a working prototype of DRS:

drs_work

bietkul (approved):

PR #1195, opened 2019-10-10
feat(web): remove react-highlight-words dependency

ShahAnuj2610:


packages/web/src/components/search/addons/Highlight.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { injectGlobal } from 'emotion';
+
+// eslint-disable-next-line no-unused-expressions
+injectGlobal`
```
jyash97: Why we are injecting it global? we can instead add it to the parent selector

packages/web/src/components/search/addons/SuggestionItem.js
```diff
@@ -19,12 +12,7 @@ const SuggestionItem = ({ currentValue, suggestion }) => {
 		// label has highest precedence
 		return typeof label === 'string' ? (
 			<div className="trim">
-				<Highlight
-					searchWords={currentValue.split(' ')}
-					textToHighlight={label}
-					autoEscape
-					highlightStyle={highlightStyle}
-				/>
+				<Highlight searchWords={currentValue.split(' ')} textToHighlight={label} />
```
bietkul: @ShahAnuj2610 What does the `autoEscape` prop do?
Did you implement it?

packages/web/src/components/search/addons/Highlight.js
```diff
@@ -0,0 +1,24 @@
...
+				),
+			}}
+		/>
+	);
+};
+
+Highlight.propTypes = {
+	searchWords: PropTypes.arrayOf(PropTypes.string),
+	textToHighlight: PropTypes.string,
+};
```
bietkul, line 43: @ShahAnuj2610 The props aren't marked as required, define the default props so the component doesn't break.

packages/web/src/styles/Input.js
```diff
@@ -146,6 +146,13 @@ const suggestions = (themePreset, theme) => css`
 		&:focus {
 			background-color: #eee;
 		}
+
```
bietkul: @ShahAnuj2610 You defined the class under `suggestions`, I don't think it'll work for `ReactiveList`.
You can create a styled component & use it.

bietkul (changes requested):

packages/web/src/styles/Input.js
ShahAnuj2610: yeah, right! But I can't use styled component since I am using `dangerouslysetInnerHtml`, so will be using emotion's `injectGlobal`. Thoughts @bietkul @jyash97?

packages/web/src/components/search/addons/Highlight.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { injectGlobal } from 'emotion';
+
+import { escapeRegExp } from '../../../utils';
+
+// eslint-disable-next-line no-unused-expressions
+injectGlobal`
```
jyash97: @bietkul  Is it good to use `injectGlobal` , instead, we can give className to the `div` and use selector for applying css.

bietkul: @ShahAnuj2610 Can you make that change?

bietkul (changes requested):

jyash97 (approved):

bietkul (approved):

PR #1197, opened 2019-10-12
feat(DropDown): add data param to render item

### LINKED ISSUE #1170: [Feature Request] Allow to use `data` in MultidropdownList and whatever
**Affected Projects**
React

**Is your feature request related to a problem? Please describe.**
I use some aggregations to get more informations rather than just labels ands counts.

However, `renderItem` doesn't give you the opportunity to use  `data` where `render` allows you that.

I could use `render` but the thing is the `renderItem` way is great for me.

So I would like to have like a 4th parameter like `(label, count, isSelected, data)`.

Thank you.

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #1170 
- [x] enhancement
- [ ] Includes tests
- [x] Documentation update: appbaseio/Docs#64 .

#### Overview of change:
pass current `item` to `renderItem` prop in `DropDown`

### DIFF
--- a/packages/vue/src/components/shared/DropDown.jsx
+++ b/packages/vue/src/components/shared/DropDown.jsx
@@ -168,6 +168,7 @@ const Dropdown = {
 															count: item.doc_count,
 															isChecked:
 																selected && this.$props.multi,
+															data: itemsToRender,
 														})
 													) : (
 														<div>
--- a/packages/vue/src/components/shared/DropDown.jsx
+++ b/packages/vue/src/components/shared/DropDown.jsx
@@ -168,7 +168,7 @@ const Dropdown = {
 															count: item.doc_count,
 															isChecked:
 																selected && this.$props.multi,
-															data: itemsToRender,
+															data: item,
 														})
 													) : (
 														<div>
--- a/packages/web/src/components/shared/Dropdown.js
+++ b/packages/web/src/components/shared/Dropdown.js
@@ -217,6 +217,7 @@ class Dropdown extends Component {
 															item[labelField],
 															item.doc_count,
 															selected && this.props.multi,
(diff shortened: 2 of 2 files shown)


packages/vue/src/components/shared/DropDown.jsx
```diff
@@ -168,6 +168,7 @@ const Dropdown = {
 															count: item.doc_count,
 															isChecked:
 																selected && this.$props.multi,
+															data: itemsToRender,
```
jyash97: Hey @tanmoyopenroot  Thanks for taking the time to fix the issue. 

It should be `item`, not `itemsToRender`.
Also, currently the `items` prop is an array of object `{ doc_count, key }` we may need to add more data to each object.

tanmoyopenroot: Done, Passing only the current item

packages/vue/src/components/shared/DropDown.jsx
```diff
@@ -168,6 +168,7 @@ const Dropdown = {
 															count: item.doc_count,
 															isChecked:
 																selected && this.$props.multi,
+															data: item,
```
jyash97: This won't work currently as the object will be `label` & `count` and we want all the data.

bietkul (approved):

jyash97:
@bietkul  I don't think we have all the value in `data`. It may only contain `key, doc_count`  for `MultiDropdownList`.

bietkul:
Cool, Reverted it since right now this PR makes doesn't make any sense.
We internally use the term query for the aggregations which only returns the `key` & `doc_count` which is already provided in `renderItem`.

PR #1211, opened 2019-10-24
chore: update readme

ShahAnuj2610:


siddharthlatest (changes requested):
@ShahAnuj2610 You should also add Searchbox here.

Order would be:
- Arc
- Searchbox (this would contain links to all Searchbox, React Searchbox, and Vue Searchbox)
- Dejavu
- Mirage
- appbase-js

siddharthlatest (approved):

Issue #1218, opened 2019-11-01
ReactiveList pagination button bug


WGriffing:
**Affected Projects**
React

**Library Version:** x.y.z
3.0.3

**Describe the bug**
The pagination buttons do not work as expected for the `ReactiveList` (or `ResultList`) component.

I think there are 2 problems:
1. When `showEndPage={false}`, the last 2 pages do not display the pagination buttons properly.
2. When `showEndPage={true}` and `pages={5}`, the last 2 pages receive dedicated buttons, such as `1 ... 3 4 ... (N-1) N`. If the number of pages is increased, the last number of pages with dedicated buttons also increases.

**To Reproduce**
https://codesandbox.io/s/reactivelist-pagination-problem-rpeij

**Problem 1:**
1. Visit sandbox link above and navigate to https://rpeij.csb.app/?SearchResult=1894 in the embedded browser pane.
1. Scroll down to pagination buttons.
1. Observe that buttons are not as expected. Instead of seeing buttons for `Prev 1890 1891 1892 1893 1894`, we see buttons `Prev 1 Next`.
1. Click `Prev` button once.
1. Observe that buttons still are not as expected.
1. Click `Prev` button once.
1. Observe that the buttons finally look normally except that page `1892` appears to be the last page visually. We already know this is not the case.
1. Click `Next` once.
1. Observe problem has re-occurred.
1. Click `Next` button once.
1. Observe you have been redirected (unexpectedly) to results page 1 rather than ending up back at page 1894.

**Problem 2:**
1. Visit sandbox link above.
1. Edit `index.js` and change to `showEndPage={true}` in the `ReactiveList`.
1. Edit `index.js` and change to `pages={10}` in the `ReactiveList`.
1. Navigate to https://rpeij.csb.app/?SearchResult=1886 in the embedded browser pane.
1. Observe that the buttons are `Prev 1 ... 1883 1884 1885 1886 1887 ... 1891 1892 1893 1894 Next`.
1. Wouldn't it be more appropriate for the buttons to be `Prev 1 ... 1882 1883 1884 1885 1886 1887 1888 1889 ... 1894 Next`?
   * Perhaps instead of 1882, it should display 1890. An even number of pages is a slightly harder case than an odd number of pages.

**Expected behavior**
* Problem 1:
  * I expect the page numbers to match the actual last page. In the sandbox, this means that page 1894 should truly be the last page.
  * The page numbers should not fail to be created on the last 2 pages.
* Problem 2:
[...]

PR #1220, opened 2019-11-04
fix(web): handle edge cases for pagination

ShahAnuj2610:
This PR attempts to fix bugs of #1218 

**What it covers:**
- When `showEndPage={false}`, the last 2 pages do not display the pagination buttons properly
- on clicking `Next` button the last pages do not render

**What it does not cover (this is more of a feature):**

when `showEndPage={true}`, does not render as per point `6`

**Testing to do:**
- [ ] `showEndPage={true}` and `totalPages <= pageSize`
- [ ] `showEndPage={false}` and `totalPages <= pageSize`
- [ ] `showEndPage={true}` and `totalPages > pageSize`
- [ ] `showEndPage={false}` and `totalPages > pageSize`
- [ ] `showEndPage={true}` and clicking on first, second, last and second_last page
- [ ] `showEndPage={false}` and clicking on first, second, last and second_last page
- [ ] `showEndPage={true}` and clicking on next, prev
- [ ] `showEndPage={false}` and clicking on next, prev
- [ ] test with combination of filter and search component

### DIFF
--- a/packages/web/src/components/result/addons/Pagination.js
+++ b/packages/web/src/components/result/addons/Pagination.js
@@ -33,7 +33,7 @@ const buildPaginationDOM = (props, position) => {
 					? start + (Math.ceil(pages / 2) - (pages % 2))
 					: totalPages + 1;
 		}
-		if (currentPage > (totalPages - pages) + 2) {
+		if (showEndPage && currentPage > (totalPages - pages) + 2) {
 			start = (totalPages - pages) + 2;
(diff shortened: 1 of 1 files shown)


ShahAnuj2610:
closing this, there are bugs for the following cases and solving them creates side-effects for other cases:

- `totalPages <= pages`
- `{showEnd=false}` => combining with filter sends wrong `from` parameter

PR #1222, opened 2019-11-06
feat(web): add aggregationField in DataSearch

ShahAnuj2610:


packages/web/src/components/list/utils.js
```diff
@@ -81,24 +81,27 @@ const getAggsQuery = (query, props) => {
 	return { ...clonedQuery, ...extractQueryFromDefaultQuery(props.defaultQuery) };
 };
 
-const getCompositeAggsQuery = (query, props, after) => {
+const getCompositeAggsQuery = (query, props, after, showTopHits = false) => {
```
bietkul: @ShahAnuj2610 Can you move the whole content of this file in `reactivecore` so we don't need to make these changes for `vue` & `react-native`.

packages/web/src/components/search/DataSearch.js
```diff
@@ -80,8 +86,11 @@ class DataSearch extends Component {
 		props.setQueryListener(props.componentId, props.onQueryChange, props.onError);
 
 		if (props.highlight) {
-			const queryOptions = DataSearch.highlightQuery(props) || {};
+			let queryOptions = DataSearch.highlightQuery(props) || {};
 			queryOptions.size = 20;
+			if (this.shouldGenerateAggsQuery) {
```
bietkul: You're repeating this block, it's better to create a method to update the query options based on the `aggregationField`.

packages/web/src/components/search/DataSearch.js
```diff
@@ -109,10 +118,13 @@ class DataSearch extends Component {
 		checkSomePropChange(
 			this.props,
 			prevProps,
-			['highlight', 'dataField', 'highlightField'],
+			['highlight', 'dataField', 'highlightField', 'aggregationField'],
 			() => {
-				const queryOptions = DataSearch.highlightQuery(this.props) || {};
+				let queryOptions = DataSearch.highlightQuery(this.props) || {};
 				queryOptions.size = 20;
+				if (this.shouldGenerateAggsQuery) {
```
bietkul: Check the above comment.

packages/web/src/components/search/DataSearch.js
```diff
@@ -998,21 +1023,34 @@ DataSearch.defaultProps = {
...
-	themePreset: state.config.themePreset,
-	isLoading: state.isLoading[props.componentId] || false,
-	error: state.error[props.componentId],
-	analytics: state.analytics,
-	config: state.config,
-	headers: state.appbaseRef.headers,
-});
+const mapStateToProps = (state, props) => {
+	const stateAggs = state.aggregations[props.componentId];
+	const computeSuggestions = () => {
```
bietkul: @ShahAnuj2610 We should not compute state in `mapStateToProps` because it can lead to performance issues. You can do this computation in `reducer` in `reactivecore`.

packages/web/src/utils/index.js
```diff
@@ -138,3 +138,15 @@ export function parseValueArray(originalArr = [], currentValue) {
 export function escapeRegExp(string) {
 	return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
 }
+
+export function parseCompAggToHits(aggFieldName, buckets = []) {
```
bietkul: Please move this function to `reactivecore`.

bietkul (changes requested):
@ShahAnuj2610 Please make the requested changes and update the rest of the PRs.

packages/web/src/components/list/utils.js
ShahAnuj2610: https://github.com/appbaseio/reactivecore/pull/48

packages/web/src/components/search/DataSearch.js
ShahAnuj2610: https://github.com/appbaseio/reactivecore/pull/48

packages/web/src/utils/index.js
ShahAnuj2610: https://github.com/appbaseio/reactivecore/pull/48

packages/web/src/components/search/DataSearch.js
```diff
@@ -80,8 +78,8 @@ class DataSearch extends Component {
 		props.setQueryListener(props.componentId, props.onQueryChange, props.onError);
 
 		if (props.highlight) {
-			const queryOptions = DataSearch.highlightQuery(props) || {};
-			queryOptions.size = 20;
+			let queryOptions = DataSearch.highlightQuery(props) || {};
+			queryOptions = { ...queryOptions, ...this.getBasicQueryOptions() };
```
bietkul: @ShahAnuj2610 The below line(`83`) is unnecessary, you can do the same here.

packages/web/src/components/search/DataSearch.js
```diff
@@ -109,10 +107,10 @@ class DataSearch extends Component {
...
 			this.props,
 			prevProps,
-			['highlight', 'dataField', 'highlightField'],
+			['highlight', 'dataField', 'highlightField', 'aggregationField'],
 			() => {
-				const queryOptions = DataSearch.highlightQuery(this.props) || {};
-				queryOptions.size = 20;
+				let queryOptions = DataSearch.highlightQuery(this.props) || {};
+				queryOptions = { ...queryOptions, ...this.getBasicQueryOptions() };
 				this.queryOptions = queryOptions;
```
bietkul: Check the above comment.

packages/web/src/components/search/DataSearch.js
```diff
@@ -184,6 +189,15 @@ class DataSearch extends Component {
 		this.props.removeComponent(this.internalComponent);
 	}
 
+	// returns size and aggs property
+	getBasicQueryOptions = () => {
+		let queryOptions = { size: 20 };
```
bietkul: @ShahAnuj2610 Define the default value for `size` prop as `20` & read it from props.

packages/web/src/components/search/DataSearch.js
```diff
@@ -998,21 +1018,30 @@ DataSearch.defaultProps = {
 	showClear: false,
 	strictSelection: false,
 	searchOperators: false,
+	size: 20,
```
bietkul, line 1021: @ShahAnuj2610 You just defined it here, but not used anywhere.

packages/web/src/components/search/DataSearch.js
```diff
@@ -998,21 +1018,30 @@ DataSearch.defaultProps = {
...
-	suggestions: state.hits[props.componentId] && state.hits[props.componentId].hits,
-	themePreset: state.config.themePreset,
-	isLoading: state.isLoading[props.componentId] || false,
-	error: state.error[props.componentId],
-	analytics: state.analytics,
-	config: state.config,
-	headers: state.appbaseRef.headers,
-});
+const mapStateToProps = (state, props) => {
+	const getSuggestionHits = () => {
```
bietkul: Again for the performance reasons don't define the function here, move it to some util file.

bietkul (changes requested):
@ShahAnuj2610 Please check the requested changes.

packages/web/src/components/search/DataSearch.js
ShahAnuj2610: @bietkul the `size` prop is for `composite` param of `aggs` => https://github.com/appbaseio/reactivecore/blob/01b710f1862dbb8101919bee5c3d6f143d4c29fb/src/utils/helper.js#L477
this `size: 20` is for the `query` param

bietkul: @ShahAnuj2610 It should support both.

ShahAnuj2610:
@bietkul have done the changes

bietkul (approved):

PR #1227, opened 2019-11-08
feat(web): add aggregationField in ReactiveComponent

ShahAnuj2610:


ShahAnuj2610:
@bietkul this can be merged

bietkul (changes requested):
@ShahAnuj2610 I think in the case of `ReactiveComponent` we should provide the `compositeAggregations` separately as we provide hits (`data` key) and `aggregations` in render & `onData` method.
There can be a use case when a user wants to get both `hits` and `aggregations` from the same query.
@siddharthlatest Can you please confirm this once?

bietkul (approved):

PR #1228, opened 2019-11-08
fix(web): aggs data key

ShahAnuj2610:


packages/web/src/components/search/CategorySearch.js
```diff
@@ -825,13 +824,14 @@ class CategorySearch extends Component {
...
 	getComponent = (downshiftProps = {}) => {
-		const { error, isLoading } = this.props;
+		const { error, isLoading, aggregationData } = this.props;
 		const { currentValue } = this.state;
 		const data = {
 			error,
 			loading: isLoading,
 			downshiftProps,
 			data: this.parsedSuggestions,
+			aggregationData,
```
bietkul, line 834: @ShahAnuj2610 Do we return `aggregationData` in `onData` method too?

packages/web/src/components/search/DataSearch.js
```diff
@@ -191,9 +190,9 @@ class DataSearch extends Component {
 	// returns size and aggs property
 	getBasicQueryOptions = () => {
 		const { aggregationField, size } = this.props;
-		let queryOptions = { size };
+		const queryOptions = { size };
 		if (aggregationField) {
-			queryOptions = getCompositeAggsQuery({ ...queryOptions }, this.props, null, true);
+			queryOptions.aggs = getCompositeAggsQuery({}, this.props, null, true).aggs;
```
bietkul, line 195: @ShahAnuj2610 Why did you change this logic?
You're reading the function value without ensuring that the function returns an object which can break the app.

bietkul (changes requested):

packages/web/src/components/search/CategorySearch.js
ShahAnuj2610: yes, but `onData` is only for `ReactiveComponent` and `ReactiveList`

bietkul: I mean `onSuggestions` in that case.

packages/web/src/components/search/DataSearch.js
ShahAnuj2610: https://github.com/appbaseio/reactivecore/blob/01b710f1862dbb8101919bee5c3d6f143d4c29fb/src/utils/helper.js#L450 => it returns an object

packages/web/src/components/search/CategorySearch.js
bietkul: @ShahAnuj2610 I checked it and found that we don't have `onData` in search components, so you can leave it for now.

PR #1243, opened 2019-11-20
feat(vue): render prop for list, dropdown and search components

ShahAnuj2610:
handles `render` prop for `List`, `Dropdown` and `Search` components

### DIFF
--- a/packages/vue/src/components/list/MultiDropdownList.jsx
+++ b/packages/vue/src/components/list/MultiDropdownList.jsx
@@ -396,6 +396,7 @@ const MultiDropdownList = {
 			const { currentValue } = this.$data;
 			const data = {
 				error: this.error,
+				loading: this.isLoading,
 				value: currentValue,
 				data: items || [],
 				handleChange: this.handleChange,
@@ -504,6 +505,7 @@ const mapStateToProps = (state, props) => ({
 		props.nestedField && state.aggregations[props.componentId]
 			? state.aggregations[props.componentId].reactivesearch_nested
 			: state.aggregations[props.componentId],
+	isLoading: state.isLoading[props.componentId],
 	selectedValue:
 		(state.selectedValues[props.componentId]
 			&& state.selectedValues[props.componentId].value)
--- a/packages/vue/src/components/list/MultiDropdownList.jsx
+++ b/packages/vue/src/components/list/MultiDropdownList.jsx
@@ -62,6 +62,7 @@ const MultiDropdownList = {
 		placeholder: VueTypes.string.def('Select values'),
 		queryFormat: VueTypes.oneOf(['and', 'or']).def('or'),
 		react: types.react,
+		render: types.func,
 		renderItem: types.func,
 		renderError: types.title,
 		transformData: types.func,
--- a/packages/vue/src/components/list/MultiDropdownList.jsx
+++ b/packages/vue/src/components/list/MultiDropdownList.jsx
@@ -259,7 +259,6 @@ const MultiDropdownList = {
 
 		handleChange(item) {
 			const { value } = this.$props;
-
 			if (value === undefined) {
 				this.setValue(item);
 			} else {
--- a/packages/vue/src/components/list/MultiDropdownList.jsx
+++ b/packages/vue/src/components/list/MultiDropdownList.jsx
@@ -391,7 +391,7 @@ const MultiDropdownList = {
 		handleLoadMore() {
 			this.updateQueryOptions(this.$props, true);
 		},
-		getComponent(items, downShiftProps = {}) {
+		getComponent(items, downshiftProps = {}) {
 			const { currentValue } = this.$data;
 			const data = {
 				error: this.error,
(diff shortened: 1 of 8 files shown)


packages/vue/src/components/list/MultiDropdownList.jsx
```diff
@@ -383,6 +392,23 @@ const MultiDropdownList = {
 		handleLoadMore() {
 			this.updateQueryOptions(this.$props, true);
 		},
+		getComponent(items, downShiftProps = {}) {
+			const { currentValue } = this.$data;
+			const data = {
+				error: this.error,
+				value: currentValue,
```
bietkul, line 399: @ShahAnuj2610 `isLoading` prop is missing here.

bietkul (changes requested):
@ShahAnuj2610 You're not exposing the `isLoading` prop in `render` props.

packages/vue/src/components/list/MultiDropdownList.jsx
ShahAnuj2610: done

packages/vue/src/components/list/MultiDropdownList.jsx
```diff
@@ -250,7 +259,6 @@ const MultiDropdownList = {
 
 		handleChange(item) {
```
bietkul, line 260: @ShahAnuj2610 You need to make `handleChange` changes here too.

packages/vue/src/components/list/SingleDropdownList.jsx
```diff
@@ -257,7 +260,7 @@ const SingleDropdownList = {
 
 		handleChange(item) {
```
bietkul, line 261: Update it too.

packages/vue/src/components/list/SingleDropdownList.jsx
```diff
@@ -1,14 +1,14 @@
 import { Actions, helper } from '@appbaseio/reactivecore';
 import VueTypes from 'vue-types';
+import { isEqual } from '@appbaseio/reactivecore/lib/utils/helper';
```
bietkul, line 3: Why did you import it i.e where you're using it?
It will have issues in watch mode if you import it like that.

packages/vue/src/components/list/SingleList.jsx
```diff
@@ -1,14 +1,14 @@
 import { Actions, helper } from '@appbaseio/reactivecore';
+import { isEqual } from '@appbaseio/reactivecore/lib/utils/helper';
```
bietkul, line 2: Same here.

packages/vue/src/components/search/DataSearch.jsx
```diff
@@ -202,6 +205,17 @@ const DataSearch = {
...
 		},
+		getComponent(downShiftProps = {}) {
+			const data = {
+				...downShiftProps,
+				loading: this.isLoading,
+				currentValue: this.$data.currentValue,
+				suggestions: this.suggestions,
+				parsedSuggestions: this.suggestionsList,
+				aggregationData: this.aggregationData,
+			};
```
bietkul, line 218: @ShahAnuj2610 Please return the same properties we used to return in `react`.
```
                         error,
			loading: isLoading,
			downshiftProps,
			data: this.parsedSuggestions,
			aggregationData,
			rawData: this.props.suggestions || [],
			value: currentValue,
			triggerClickAnalytics: this.triggerClickAnalytics,
```

bietkul (changes requested):
@ShahAnuj2610 Please sync the code with the latest.
Eslint and prettier will only work when you follow these steps:
1. Close vscode
2. Open the `packages/vue` folder

packages/vue/src/components/list/MultiDropdownList.jsx
ShahAnuj2610: we don't need here, it directly accepts `item` as a param rather than `event`

packages/vue/src/components/list/SingleDropdownList.jsx
ShahAnuj2610: we don't need here, it directly accepts `item` as a param rather than `event`

packages/vue/src/components/list/SingleDropdownList.jsx
ShahAnuj2610: I didn't import it, re-order it to solve `eslint` warning of `absolute imports should be before relative imports`

packages/vue/src/components/list/SingleList.jsx
ShahAnuj2610: I didn't import it, re-order it to solve `eslint` warning of `absolute imports should be before relative imports`

packages/vue/src/components/search/DataSearch.jsx
ShahAnuj2610: I deliberately did that, since `renderAllSuggestions` was behaving the same way and not to break existing user behaviour

bietkul: @ShahAnuj2610 Please make those changes, we can have breaking changes.

ShahAnuj2610: ok

ShahAnuj2610:
@bietkul have done the changes

bietkul (approved):

PR #1263, opened 2019-12-02
fix(web): remove searchStateHeader ReactiveBase.js

ShahAnuj2610:


bietkul:
@ShahAnuj2610 We don't need to remove it, we'll do that in 4.0.

PR #1266, opened 2019-12-03
fix(vue): style of DataSearch clear icon

ShahAnuj2610:
before: 
after:


ShahAnuj2610:
@jyash97 can you once review this?

jyash97 (approved):

PR #1267, opened 2019-12-03
fix(vue): remove defaultSelected from list

jyash97:


bietkul:
@jyash97 I fixed it, this PR is not needed.

PR #1284, opened 2019-12-16
fix(web): remove showing filters for pagination

ShahAnuj2610:


packages/web/src/components/basic/SelectedFilters.js
```diff
@@ -146,18 +150,21 @@ SelectedFilters.propTypes = {
...
 	clearAllLabel: 'Clear All',
 	showClearAll: true,
 	style: {},
+	componentProps: {},
 };
 
 const mapStateToProps = state => ({
 	components: state.components,
 	selectedValues: state.selectedValues,
+	componentProps: state.props,
```
lakhansamani, line 167: @ShahAnuj2610 can you confirm once, if this component is re-rendering on any prop change in component. if so this needs to be optimized.

ShahAnuj2610: sure

lakhansamani (approved):

PR #1286, opened 2019-12-17
fix(web): make size field optional for ReactiveComponent.d.ts

### LINKED ISSUE #1281: ReactiveComponent size prop
**Affected Projects**
React

**Library Version:** 3.1.1

**Describe the bug**
The documentation does not have any reference to a size prop for the ReactiveComponent. However in the type definitions it requires a number. 

```
export interface ReactiveComponentProps {
	children?: (...args: any[]) => any;
	render?: (...args: any[]) => any;
	componentId: string;
	aggregationField?: string;
	size: number;
	defaultQuery?: (...args: any[]) => any;
	customQuery?: (...args: any[]) => any;
	defaultValue?: any;
	value?: any;
	filterLabel?: string;
	onData?: (...args: any[]) => any;
	onError?: (...args: any[]) => any;
	react?: types.react;
	showFilter?: boolean;
	onQueryChange?: (...args: any[]) => any;
	URLParams?: boolean;
}
```
I'm not exactly sure whether the documentation or the type definitions are wrong. What is the requirement for?

I am happy to raise a PR for either case. Just need a bit of background info.

The propTypes also contradict this. They expect a number, but it is not required and a default is set to 20.

**To Reproduce**

1. Create a typescript project
2. Add a ReactiveComponent with a size prop

**Expected behavior**
No type error should be shown

ShahAnuj2610:
fixes #1281

### DIFF
--- a/packages/web/src/components/basic/ReactiveComponent.d.ts
+++ b/packages/web/src/components/basic/ReactiveComponent.d.ts
@@ -7,7 +7,7 @@ export interface ReactiveComponentProps {
 	render?: (...args: any[]) => any;
 	componentId: string;
 	aggregationField?: string;
-	size: number;
+	size?: number;
 	defaultQuery?: (...args: any[]) => any;
 	customQuery?: (...args: any[]) => any;
 	defaultValue?: any;


packages/web/src/components/basic/ReactiveComponent.d.ts
```diff
@@ -7,7 +7,7 @@ export interface ReactiveComponentProps {
 	render?: (...args: any[]) => any;
 	componentId: string;
 	aggregationField?: string;
-	size: number;
+	size?: number;
```
bietkul, line 10: @ShahAnuj2610 You have to use this `size` prop in the ReactiveComponent, if not using then remove it.

bietkul (changes requested):

packages/web/src/components/basic/ReactiveComponent.d.ts
ShahAnuj2610: @bietkul it's being used -> https://github.com/appbaseio/reactivecore/blob/3d4d01163941ec89261982ab2e180adffc4862c5/src/utils/helper.js#L454

bietkul (approved):

PR #1295, opened 2019-12-23
feat(web): showClearAll as enum in SelectedFilters.js

ShahAnuj2610:
dependent on: https://github.com/appbaseio/reactivecore/pull/54

new behaviour:
1. when `showClearAll=always`: show filters even when `showFilter={false}` for a particular component. Based on whether value is `selected` or not.
2. when `showClearAll=default`: show filters only when `showFilter={true}` for a particular component
3. when `showClearAll=never`: never show clear all button
3. when `showClearAll=true`: identical to `showClearAll=always` (for backward compatiblity)
4. when `showClearAll=false`: identical to `showClearAll=never` (for backward compatiblity)

### DIFF
--- a/packages/web/src/components/basic/SelectedFilters.js
+++ b/packages/web/src/components/basic/SelectedFilters.js
@@ -11,6 +11,19 @@ import Title from '../../styles/Title';
 import { connect } from '../../utils';
 
 class SelectedFilters extends Component {
+	constructor(props) {
+		super(props);
+		this.extracted(props);
+	}
+
+	extracted(props) {
+		if (props.showClearAll === true) {
+			this._showClearAll = 'always';
+		} else {
+			this._showClearAll = props.showClearAll === false ? 'never' : props.showClearAll;
+		}
+	}
+
 	componentDidUpdate = () => {
 		if (this.props.onChange) {
 			this.props.onChange(this.props.selectedValues);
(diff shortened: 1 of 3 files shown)


packages/web/src/types.ts
```diff
@@ -143,3 +143,5 @@ export interface analyticsConfig {
 	userId: string;
 	customEvents: object;
 }
+
+export type showClearAll = 'never' | 'always' | 'default' | true | false;
```
bietkul, line 147: @ShahAnuj2610 Can you use constants here?

packages/web/src/components/basic/SelectedFilters.js
```diff
@@ -144,7 +164,7 @@ SelectedFilters.propTypes = {
 	className: types.string,
 	clearAllLabel: types.title,
 	innerClass: types.style,
-	showClearAll: types.bool,
+	showClearAll: types.showClearAll,
```
bietkul, line 167: @ShahAnuj2610 What is the default value of it?

bietkul (changes requested):

packages/web/src/components/basic/SelectedFilters.js
ShahAnuj2610: true

packages/web/src/types.ts
ShahAnuj2610: it's a TS file, not able to import it

bietkul (approved):

packages/web/src/components/basic/SelectedFilters.js
ShahAnuj2610: https://github.com/appbaseio/reactivesearch/blob/08b63715172c8c29c194d158b66d2302762d9b24/packages/web/src/components/basic/SelectedFilters.js#L180

PR #1298, opened 2019-12-24
fix(web): daterange for different time-zones

### LINKED ISSUE #1297: DateRange (with URLparams={true}) shifts every reload
**Affected Projects**
React

**Library Version:**
3.1.3

**Describe the bug**
While using a DateRange component with `URLParams={true}` and a date range set by the user, reloading the page (or copying/pasting URL to new tab) causes the selected date range to shift. The shift amount is dependent on your timezone as compared to UTC, but for ESD/EST the shift is -1 day for every reload.

**To Reproduce**
https://codesandbox.io/s/reactivesearch-daterange-shift-bug-89lpg (forked from DateRange embedded example)
1. Enter a date range (i.e. 2017-05-01 to 2017-05-31).
2. In the embedded browser "window", click the "Reload" button.
3. Observe that the selected values are now "2017-04-30" and "2016-05-30".
4. Click "Reload" from step 2 again.
5. Observe that the selected values have decremented again.

Note: If your timezone offset is +0, you will not observe the bug.

**Expected behavior**
I expect the URL parameters to stay fixed so that if I share my URL with another user, they will end up doing the same search as I was doing. Similarly, I do not expect refreshing my browser to alter my URL parameters.

**Screenshots**
I can add screenshots if anything is unclear, but I think the sandbox reproduces the problem well.

**Desktop (please complete the following information):**
 - OS: Ubuntu 19.10
 - Browser: Firefox
 - Version: 71.0

**Additional context**
The problem occurs in packages `/web/src/components/date/DateRange.js`, lines 42-44:
```javascript
  ...
  currentDate = {
    start: new Date(props.selectedValue[0]),
    end: new Date(props.selectedValue[1]),
  };
  ...
```
This `Date` constructor is assuming the input is in UTC and then shifting to my local timezone. You can verify this by logging the value of `props.selectedValue` and comparing it to the value of `currentDate.start` and `currentDate.end`.

I think (one possible way) the problem could be resolved is by using a different `Date` constructor that doesn't assume the input is in UTC. Example solution:
```javascript
  ...
  var startArr = props.selectedValue[0].split('-');
  var endArr = props.selectedValue[1].split('-');
  currentDate = {
    start: new Date(startArr[0], startArr[1]-1, startArr[2]),
    end: new Date(endArr[0], endArr[1]-1, endArr[2]),
  };
...
```

I believe the URL parameter format is always going to be 'yyyy-MM-dd' due to line 139 in the same file (`return xdate.valid() ? xdate.toString('yyyy-MM-dd') : '';`) so we can split the input on hyphens and make use of the `Date` constructor that assumes the input is supplied in the browser's timezone.

ShahAnuj2610:
fixes #1297 

How to test: open terminal and paste following commands to open browser with different time-zone
```
mkdir ~/chrome-profile
TZ='US/Pacific' open -na "Google Chrome" --args "--user-data-dir=$HOME/chrome-profile"
```

The URL with `URLParams` should render the same date in both your `local` and different `timezoned` browser

### DIFF
--- a/packages/web/src/components/date/DateRange.js
+++ b/packages/web/src/components/date/DateRange.js
@@ -40,14 +40,14 @@ class DateRange extends Component {
 		if (props.selectedValue) {
 			if (Array.isArray(props.selectedValue)) {
 				currentDate = {
-					start: new Date(props.selectedValue[0]),
-					end: new Date(props.selectedValue[1]),
+					start: new XDate(props.selectedValue[0])[0],
+					end: new XDate(props.selectedValue[1])[0],
 				};
 			} else {
 				const { start, end } = props.selectedValue;
 				currentDate = {
-					start: new Date(start),
-					end: new Date(end),
+					start: new XDate(start)[0],
+					end: new XDate(end)[0],
 				};
 			}
 		}


bietkul (approved):

davidkong0987:
Did this issue reintroduce itself when xdate was changed back to dayjs?

davidkong0987:
> Did this issue reintroduce itself when xdate was changed back to dayjs?

related issue https://github.com/appbaseio/reactivesearch/issues/1297

PR #1302, opened 2019-12-26
fix(web): promoted data key

ShahAnuj2610:


bietkul (changes requested):
@ShahAnuj2610 Change in `StateProvider` too.

ShahAnuj2610:
> @ShahAnuj2610 Change in `StateProvider` too.

It's done in the corresponding `reactivecore`

bietkul:
> > @ShahAnuj2610 Change in `StateProvider` too.
> 
> It's done in the corresponding `reactivecore`

@ShahAnuj2610 I don't think it'll work in that way, can you verify?

ShahAnuj2610:
> > > @ShahAnuj2610 Change in `StateProvider` too.
> > 
> > 
> > It's done in the corresponding `reactivecore`
> 
> @ShahAnuj2610 I don't think it'll work in that way, can you verify?

it works

PR #1309, opened 2020-01-01
refactor(web): styles for DataSearch and CategorySearch

ShahAnuj2610:
- [x]  Fix icon styles
    - [x]  Alignment is off

        

    - [x]  Icon colors are different
    - [x]  There's an overlap issue: See here - [https://codesandbox.io/s/competent-hamilton-2f2jy](https://codesandbox.io/s/competent-hamilton-2f2jy)
- [x]  Fix max-height for suggestions to show min 10 rows (this also applies to Searchbox libraries). We currently show a scroll.
- [x]  increase `max-width` to avoid scroll


jyash97 (changes requested):
- Check clear icon color: ( both icon dont have same color )
image

- When the mic is active we switched to black shade color but I guess it should be set to primary color
image

- The mic icon looks too small. Check this layout ( doesn't look good may be due to different width ) Thoughts @siddharthlatest 
image

jyash97 (commented):
image

Can we have property `text-overflow: ellipsis` on suggestion ?

jyash97 (approved):

PR #1311, opened 2020-01-02
refactor(vue): styles for DataSearch

ShahAnuj2610:
- [x]  Fix icon styles
    - [x]  Alignment is off
    - [x]  There's an overlap issue
- [x]  Fix max-height for suggestions to show min 10 rows (this also applies to Searchbox libraries). We currently show a scroll.
- [x]  increase `max-width` to avoid scroll


jyash97 (commented):
In this PR we aren't fixing the color for svg or we have different PR for that

ShahAnuj2610:
> In this PR we aren't fixing the color for svg or we have different PR for that

we don't have `Mic` component here

jyash97:
So in vue we don't have voice search ?

ShahAnuj2610:
> So in vue we don't have voice search ?

no

PR #1316, opened 2020-01-07
feat(maps): use geocode client side APIs

ShahAnuj2610:
### How to test

Run any of the map examples on port `8001`

### DIFF
--- a/packages/maps/src/components/basic/GeoCode.js
+++ b/packages/maps/src/components/basic/GeoCode.js
@@ -1,0 +1,46 @@
+import { Component } from 'react';
+
+class GeoCode extends Component {
+	getUserLocation() {
+		navigator.geolocation.getCurrentPosition((location) => {
+			const coordinatesObj = {
+				lat: location.coords.latitude,
+				lng: location.coords.longitude,
+			};
+
+			this.geocoder.geocode({ location: coordinatesObj }, (results, status) => {
+				if (status === 'OK') {
+					if (Array.isArray(results) && results.length) {
+						const userLocation = results[0].formatted_address;
+						this.setState({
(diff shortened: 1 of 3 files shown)


packages/maps/src/components/basic/GeoCode.js
```diff
@@ -0,0 +1,46 @@
+import { Component } from 'react';
+
+class GeoCode extends Component {
+	getUserLocation() {
+		navigator.geolocation.getCurrentPosition((location) => {
```
bietkul, line 5: @ShahAnuj2610 What's the need for this kind of pattern, can you explain?

packages/maps/src/components/basic/GeoCode.js
```diff
@@ -0,0 +1,46 @@
...
+				lng: location.coords.longitude,
+			};
+
+			this.geocoder.geocode({ location: coordinatesObj }, (results, status) => {
+				if (status === 'OK') {
+					if (Array.isArray(results) && results.length) {
+						const userLocation = results[0].formatted_address;
+						this.setState({
+							// eslint-disable-next-line react/no-unused-state
+							userLocation,
```
bietkul, line 17: @ShahAnuj2610 Why did you define state if not using it?

bietkul (changes requested):

bietkul (approved):

packages/maps/src/components/basic/GeoCode.js
ShahAnuj2610: `setState` gets called in the context of `this` of parent component like `GeoDistanceSlider` and `GeoDistanceDropdown`

PR #1317, opened 2020-01-07
fix(web): re-render of Reactivelist due to analytics prop

jyash97:
Dependent on: https://github.com/appbaseio/reactivecore/pull/60

### DIFF
--- a/packages/web/src/components/result/ReactiveList.js
+++ b/packages/web/src/components/result/ReactiveList.js
@@ -8,6 +8,7 @@ import {
 	setQueryOptions,
 	updateQuery,
 	loadMore,
+	recordResultClick,
 	setValue,
 	setQueryListener,
 	setComponentProps,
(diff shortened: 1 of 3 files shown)


packages/web/src/components/result/ReactiveList.js
```diff
@@ -882,18 +863,16 @@ const mapStateToProps = (state, props) => ({
 			&& state.selectedValues[props.componentId].value - 1)
 		|| -1,
 	hits: state.hits[props.componentId] && state.hits[props.componentId].hits,
-	aggregationData: state.compositeAggregations[props.componentId] || [],
+	aggregationData: state.compositeAggregations[props.componentId],
```
bietkul, line 866: @jyash97 What's the need of this change?

packages/web/src/components/search/DataSearch.js
```diff
@@ -1001,13 +979,11 @@ const mapStateToProps = (state, props) => ({
 			&& state.selectedValues[props.componentId].value)
 		|| null,
 	suggestions: state.hits[props.componentId] && state.hits[props.componentId].hits,
-	aggregationData: state.compositeAggregations[props.componentId] || [],
+	aggregationData: state.compositeAggregations[props.componentId],
```
bietkul, line 982: @jyash97 Same here

bietkul (changes requested):

packages/web/src/components/search/DataSearch.js
jyash97: This is causing an extra re-render because of the fallback.

bietkul: @jyash97 I'm not sure if @ShahAnuj2610 handled it well.
@ShahAnuj2610 Can you confirm if this change wouldn't break the things?

PR #1318, opened 2020-01-07
fix(web): aggregationData return type for render prop

ShahAnuj2610:


bietkul (approved):

packages/web/src/components/search/CategorySearch.js
```diff
@@ -810,7 +810,7 @@ class CategorySearch extends Component {
 			downshiftProps,
 			data: this.parsedSuggestions,
 			promotedData: promotedResults,
-			aggregationData,
+			aggregationData: aggregationData || [],
```
jyash97, line 813: Shouldn't this be already handled using `defaultProp` ?

PR #1332, opened 2020-01-17
refactor: refactor StateProvider component to hooks

meetzaveri:


packages/web/src/components/basic/StateProvider.js
```diff
@@ -43,39 +47,60 @@ const filterByKeys = (state, allowedKeys) =>
...
+	]);
+
+	// prevent extra re-render on change of every respective prop if value remains same towards comparision
+	const view = useMemo(
+		() => <React.Fragment>{getComponent({ searchState }, mergeProps)}</React.Fragment>,
+		[shouldUpdate],
+	);
+
+	// // Listen for state change and trigger the callback for useEffect
+	React.useEffect(() => {
```
lakhansamani: You can use `useEffect` directly as you have imported

packages/web/src/components/basic/StateProvider.js
```diff
@@ -43,39 +47,60 @@ const filterByKeys = (state, allowedKeys) =>
...
+		// console.log('Redux state change', prevState);
```
lakhansamani: remove this comment

packages/web/src/components/basic/StateProvider.js
```diff
@@ -43,39 +47,60 @@ const filterByKeys = (state, allowedKeys) =>
...
+
+		const { onChange } = props;
+		// call onChange() if it has been defined in parent
+		onChange && onChange(prevState.searchState, this.state.searchState);
```
lakhansamani: `this.state.searchState` is wrongly used over here. Since it is a functional component it should be only `searchState`

lakhansamani: Since we are changing state over here, we should add a memory leak check.

example:

```
useEffect(() => {
   let isMounted = true;
   
  if (isMounted) {
    // on chnage logic
  }
  
 return function cleanUp() {
   isMounted = false;
 }
}, [])

```

This will help us prevent any memory leak when a state is change is executed on unmounted component.

packages/web/src/components/basic/StateProvider.js
```diff
@@ -95,17 +120,53 @@ StateProvider.propTypes = {
...
-	queryLog: filterByComponentIds(state.queryLog, props),
-	dependencyTree: filterByComponentIds(state.dependencyTree, props),
-	componentProps: filterByComponentIds(state.props, props),
-	hits: filterByComponentIds(state.hits, props),
-	aggregations: filterByComponentIds(state.aggregations, props),
-	isLoading: filterByComponentIds(state.isLoading, props),
-	error: filterByComponentIds(state.error, props),
-	promotedResults: filterByComponentIds(state.promotedResults, props),
-});
+/* Old State Provider component */
```
lakhansamani: You can remove old class component

lakhansamani (changes requested):

packages/web/src/components/basic/StateProvider.js
meetzaveri: On it!

packages/web/src/components/basic/StateProvider.js
meetzaveri: Sure

packages/web/src/components/basic/StateProvider.js
meetzaveri: So should replace it with this example code or add it below the onchange useeffect hook?

meetzaveri:
@lakhansamani  I have made changes acc. to your request. Could you review it?

PR #1341, opened 2020-01-30
fix(web): tag cloud example

ShahAnuj2610:
the demo was breaking here: https://codesandbox.io/s/github/appbaseio/reactivesearch/tree/next/packages/web/examples/TagCloud?from-embed

### DIFF
--- a/packages/web/examples/TagCloud/src/index.js
+++ b/packages/web/examples/TagCloud/src/index.js
@@ -65,7 +65,7 @@ const Main = () => (
 												</div>
 												<div className="flex wrap meetup-topics">
 													{data.group
-														&& data.group.group_topics
+														&& (data.group.group_topics || [])
 															.slice(0, 4)
 															.map(tag => (
 																<div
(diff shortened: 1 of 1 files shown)


packages/web/examples/TagCloud/src/index.js
```diff
@@ -61,16 +64,17 @@ const Main = () => (
...
-														.map(tag => (
-															<div
-																className="meetup-topic"
-																key={tag.topic_name}
-															>
-																{tag.topic_name}
-															</div>
-														))}
+													{data.group
+														&& data.group.group_topics
```
jyash97: `data.group && data.group.group_topics &&  ...`

ShahAnuj2610: It's `data.group && data.group.group_topics.slice`

jyash97: We can add check for `groupd_topics` as well. Just to be on safe side

ShahAnuj2610: done

jyash97 (approved):

Issue #1338, opened 2020-01-24
Using DataSearch with showVoiceSearch on SSR failing window is not defined


aleplusplus:
**Affected Projects**
React

**Library Version:** 3.2.4

**Describe the bug**
Using DataSearch with showVoiceSearch on SSR failing window is not defined

**To Reproduce**

Using NextJS with DataSearch and the prop `showVoiceSearch=true`

**Expected behavior**

**Screenshots**

<img width="1305" alt="Screen Shot 2020-01-24 at 11 09 00 AM" src="https://user-images.githubusercontent.com/6319838/73083896-fc2dcd80-3e99-11ea-9f47-3d3e894f5ae3.png">

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser Chrome
 - Version Version 79.0.3945.117 (Official Build) (64-bit)

**Additional context**

ShahAnuj2610:
@aleplusplus you need to tweak code a little bit. Here's a working example for you: https://codesandbox.io/s/elated-flower-rm2wo

You can do this using `Wrapper` component:
```
const Wrapper = props => {
  return <DataSearch {...props} />;
};

// class
state = { mount: false };
componentDidMount() {
  this.setState({ mount: true });
}

//render
{this.state.mount && <Wrapper {...components.datasearch} />}
```

PR #1345, opened 2020-01-31
fix(web): window undefined for ssr example

ShahAnuj2610:
fixes example for SSR referred here: #1338

### DIFF
--- a/packages/web/examples/ssr/pages/index.js
+++ b/packages/web/examples/ssr/pages/index.js
@@ -32,6 +32,7 @@ const components = {
 		className: 'search',
 		highlight: true,
 		URLParams: true,
+		showVoiceSearch: true,
 	},
 	numberbox: {
 		componentId: 'GuestSensor',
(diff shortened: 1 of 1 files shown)


packages/web/examples/ssr/pages/index.js
```diff
@@ -101,7 +102,15 @@ const components = {
...
 
+const Wrapper = props => {
+	return <DataSearch {...props} />;
+};
+
 export default class Main extends Component {
+	state = { mount: false };
+	componentDidMount() {
+		this.setState({ mount: true });
+	}
```
jyash97, line 113: Shouldn't this be a part of library? As on SSR this will break for sure because `window` will be undefined. Any specific reason we are doing this on example ? 

cc @lakhansamani @bietkul What do you guys think ?

lakhansamani: Yes, this should be part of the library. We in fact do something similar in ReactiveMaps as we cannot have maps rendered on the server. We should have checked there itself like,

```
if (window) {
   ....
}

```

ShahAnuj2610: OK

ShahAnuj2610:
fixed here: https://github.com/appbaseio/reactivesearch/pull/1351

Issue #1205, opened 2019-10-21
DynamicRangeSlider not rendering when minimum value is 0


JakeShelley1:
**Affected Projects**
React

**Library Version:** x.y.z
v3.0.3

**Describe the bug**
I have a field of datatype long called `rating_count`. When I use that as the `dataField` for my `DynamicRangeSlider` component the component does not render. 

The network request goes out and returns the proper response with the min and max of the field. When I update my index to make all `rating_count` values greater than zero, the `DynamicRangeSlider` renders properly.

**To Reproduce**
1. Create an index with a `long` field
2. Add a doc to that index where your chosen field is `0`
3. Setup `DynamicRangeSlider` as it says in the docs
4. Your slider will not render

**Expected behavior**
`DynamicRangeSlider` should render with the minimum value of 0

**Desktop (please complete the following information):**
Chrome v77.0.3865.120

**Additional context**
I'm not sure if this is expected behavior. Looking at the example (with the books) the minimum there is 1, so maybe this should be classified as an enhancement?

JakeShelley1:
I've continued to dig on this with no luck. For more context, here's the query that my DynamicRangeSlider makes initially:
```
{
  "query": {
    "match_all":{}
  },
  "aggs": {
    "min": {
      "min": {
        "field": "review_count" 
      }
    },
    "max": {
      "max": {
        "field": "review_count"
      }
    }
  }
}
```

The `aggregations` field this returns looks like this
```
{
  "aggregations" : {
    "min" : {
      "value" : 0.0
    },
    "max" : {
      "value" : 12071.0
    }
  }
```
After removing all 0 values from the `review_count` field I'm searching, I tried again and the format of the result looks identical (except the values being 1.0 and 27.0 respectively), but the component still doesn't render.

rp4rk:
Also experiencing this, exact same recreation steps as Jake.

rp4rk:
https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/range/DynamicRangeSlider.js#L69

I believe this line is the problem, `this.props.range.start` will be coerced to false and cause the block to never be entered, meaning `this.state.range` is never set resulting in the component not rendering.

Issue #1054, opened 2019-07-15
Clearing filters when a DynamicRangeSlider is present doesn't return all results to List


stefanlesik:
Affected Projects
React

Library Version:
3.0.0-rc.20

Describe the bug
Clearing filters when a DynamicRangeSlider is present doesn't return all results to List

To Reproduce
Sandbox:
https://codesandbox.io/embed/eager-leftpad-mhrp4

1. Enter search term "The Illearth War". Hit Enter.
2. Click 'Clear All'

Expected behavior
Should be returned to initial state with all products in view

Macbook
Chrome

ebytton:
is there any workaround for this issue? I am able to reproduce the problem with 3.0.0-rc.23

Thanks

ebytton:
This workaround worked for me. 

` <DynamicRangeSlider
          dataField="ratings_count"
          componentId="BookSensor"
          showFilter={false}
          beforeValueChange={
            function (value) {
              return new Promise((resolve, reject) => {
                if (value.start === value.end) {
                  reject()
                } else {
                  resolve()
                }
              })
            }
          }
          react={{
            and: ["title"]
          }}
        />`
Do not update the range if start == end.

Issue #739, opened 2018-12-28
DynamicRangeSlider breaks search when no results found


bietkul:
**Issue Type:**
Bug

**Platform:**
Web

**Description:**
DynamicRangeSlider adds an unnecessary filter with min & max value as `0` when there no results present for the search keyword.
It breaks the search completely until I refresh the page.

**Screenshots:**
![alt text](https://i.imgur.com/mjhnGZk.png)

**Minimal reproduction of the problem with instructions:**
Use DynamicRangeSlider component & try to search something for which there no results present.

**Reactivesearch version:** x.y.z

**Browser:** [all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

**Anything else:**

janus-reith:
Any news regarding this issue?

PR #1355, opened 2020-02-06
fix(web): DynamicRangeSlider Fix for Start Value of Zero

rp4rk:
[Relevant Issue](https://github.com/appbaseio/reactivesearch/issues/1205)

# Description
Adds a numerical check for `range.start`, previously having a zero value for `range.start` resulted in the component not being rendered as values never propagated into the component state due to this check erroneously being false.

### DIFF
--- a/packages/web/src/components/range/DynamicRangeSlider.js
+++ b/packages/web/src/components/range/DynamicRangeSlider.js
@@ -64,9 +64,9 @@ class DynamicRangeSlider extends Component {
 			this.props.updateComponentProps(this.props.componentId, this.props);
 		});
 		if (
-			!isEqual(this.props.range, prevProps.range)
-			&& this.props.range
-			&& this.props.range.start
+			!isEqual(this.props.range, prevProps.range) &&
+			this.props.range &&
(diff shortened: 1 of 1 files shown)


ShahAnuj2610:
@rp4rk thanks for the PR. can you once check these two scenarios?

- https://github.com/appbaseio/reactivesearch/issues/1054
- https://github.com/appbaseio/reactivesearch/issues/739

rp4rk:
* https://github.com/appbaseio/reactivesearch/issues/1054
Doesn't seem to be an issue anymore, can't recreate with the fix.

* https://github.com/appbaseio/reactivesearch/issues/739
Does seem to have regressed, will need more investigation

rp4rk:
I've added an upper bound check, I don't think having the start and end both set to `0` is a valid state. This should fix the regression with #739, what do you think?

ShahAnuj2610 (approved):

rp4rk:
Excellent, how do I address the failing label check?

ShahAnuj2610:
> Excellent, how do I address the failing label check?

no need to, you can ignore that.

Issue #1328, opened 2020-01-15
Please remove the typo error in MultiDataList js code for customQuery to work


e-gaulue:
https://github.com/appbaseio/reactivesearch/blob/640f970330509d7a85bf40b1125ddbe75e88f2b8/packages/web/src/components/list/MultiDataList.js#L289

The call just need single parentheses. Maybe somme other places in the code are concerned.

Regards,

e-gaulue:
It's on the way: #1359

alexfgo:
The same problem appears to be happening on defaultQuery property.
At this moment it seems impossible to provide any custom function on both properties.

Regards

PR #1359, opened 2020-02-06
Fix a typo bugg. customQuery is waiting for 2 vars, not just the last…

e-gaulue:
… one.


crab86:
Why is this closed? On #1328 it says the fix is on the way, but #1359 (this one) which should provide the fix is without any commit but closed?

Issue #1112, opened 2019-08-22
Query for Multilist Filter is altered on load


hansendx:
**Affected Projects**
Vue.JS

**Library Version:** 1.0.0-beta.18

**Describe the bug**

I am using a Multilist filter.
When setting  `:URLParams="true"`, something like `Filter=["thing1","thing2"]` is added to the query part of the url. When reloading the page or taking the url and freshly loading it
`Filter=["thing1","thing2"]` is changed to `Filter=["thing1,thing2"]`.
In addition the results are then also filtered using `"thing1,thing2"` which probably eliminates all results in most cases. 
This happens for all lists longer than one. All occurrences of `","` are replaced by just a `,`.

**To Reproduce**
I have set `:URLParams="true"` in a forked codesandbox
https://codesandbox.io/embed/multi-list-2yure

To reproduce, one has to select multiple filters and reload page in the demo browser.

**Expected behavior**

I would expect `Filter=["thing1","thing2"]` to be kept as `Filter=["thing1","thing2"]`.
The react project seems to handle this correctly:
https://codesandbox.io/embed/multilist-3io0w

I cannot reproduce the bug there.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04, Windows 7
 - Browser: Chrome, Firefox
 - Version: 75.0.3770.142, 68.0.2

**Additional context**

I tried to follow the query through a debugger to identify the problem.
The furthest I got was that the [selectedValues](https://github.com/appbaseio/reactivesearch/blob/8f5ff252d487ed486a68eb619ba026c0b5b922f6/packages/vue/src/components/URLParamsProvider.jsx#L53) function inside the URLParamsProvider component is called with `oldVal` and `newVal` by a vue [watcher function](https://github.com/vuejs/vue/blob/e90cc60c4718a69e2c919275a999b7370141f3bf/dist/vue.runtime.esm.js#L4546).
There `oldVal` still contains the correct `["thing1","thing2"]` string while `newVal` contains `["thing1,thing2"]`.

hansendx:
This problem persist in the 1.0.0 stable release.

EnnioSimoes:
I have the same problem

MrCode1997:
I have the same problem too.

hansendx:
This is still a problem in version 1.7.0
https://codesandbox.io/s/sleepy-cdn-ffruv?file=/src/App.vue
[...]

PR #1364, opened 2020-02-13
fix(#1112): fix multi parameters for multilist component

MrCode1997:
- [x] That change is need bracuse of bug that change the urlparamters in the URL when the props :URLParams is set to "true" on MultiList component. So with this change we are able to use MultiList component with url params with no bug.

Thanks All.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
 	"private": true,
 	"name": "reactivesearch-packages",
+	"version": "1.0.0",
 	"scripts": {
 		"dev:web": "concurrently --kill-others -n \"CORE\",\"WEB\",\"MAPS\", -c \"cyan\",\"yellow\",\"green\", \"npm start --prefix packages/reactivecore\" \"npm run watch --prefix packages/web\" \"npm run watch --prefix packages/maps\"",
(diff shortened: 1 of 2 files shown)


bietkul:
@MrCode1997 I'm not sure what're you doing here, Can you please first explain the issue with a reproducible example? A code sandbox link will be much helpful.
Thanks

package.json
```diff
@@ -1,6 +1,7 @@
 {
 	"private": true,
 	"name": "reactivesearch-packages",
+	"version": "1.0.0",
```
bietkul, line 4: @MrCode1997 We don't need this change.

bietkul (changes requested):

package.json
MrCode1997: @bietkul Ok, perfect probably was a refuse. i can remove it.

MrCode1997:
> @MrCode1997 I'm not sure what're you doing here, Can you please first explain the issue with a reproducible example? A code sandbox link will be much helpful.
> Thanks

Hi @bietkul there is an open bug: https://github.com/appbaseio/reactivesearch/issues/1112
So i solved this bug. This is the only thing that i do.

gpassarelli:
Hi Guys, this PR solded the issue #1112 for any, can you guys by any change approve this PR?

bietkul:
This issue has been fixed in https://github.com/appbaseio/reactivesearch/pull/1602

PR #1365, opened 2020-02-17
feat(reactivemap): better support for OpenStreetMap (zoom, refresh...)

e-gaulue:
In fact we can get a kind of "Google" mapRef with react-leaflet. So I followed the same way as ReactiveGoogleMap with a mapRef state.
```
ref={(elem) => {
	if (!this.state.mapRef) {
		this.setState({
			mapRef: elem.leafletElement,
		});
	}
}}
```

I added a mapService string prop to both Reactive(Google|OpenStreet)Map tools. They get a default value 'GoogleMap' or 'OpenStreetMap' during loading. We need this to have specific treatment in ReactiveMap component. Maybe we could get this information directly from mapRef.

I changed mapRef existence test `!!this.props.mapRef` (not lint compliant)  to function existence test in this mapRef "object". Bye the way, `!!this.props.mapRef` gave me unexplained behaviour with OpenStreetMap whereas function test works really well. 

I added onZoomEnd event to handle zoom change for OpenStreetMap (main reason I did all this).

I added `viewport={{}}` due to this https://react-leaflet.js.org/docs/en/components "Manipulating viewport" chapter. Without it, sometimes, changing center and zoom was not handle.

### DIFF
--- a/packages/maps/src/components/result/ReactiveGoogleMap.js
+++ b/packages/maps/src/components/result/ReactiveGoogleMap.js
@@ -219,6 +219,7 @@ ReactiveGoogleMap.propTypes = {
 	renderMap: types.func,
 	updaterKey: types.number,
 	mapRef: types.any, // eslint-disable-line
+	mapService: types.string,
 };
 
 ReactiveGoogleMap.defaultProps = {
@@ -242,6 +243,7 @@ ReactiveGoogleMap.defaultProps = {
 	showMarkerClusters: true,
 	unit: 'mi',
 	defaultRadius: 100,
+	mapService: 'GoogleMap',
 };
 
 export default ReactiveGoogleMap;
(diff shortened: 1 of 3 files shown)


packages/maps/src/components/result/ReactiveGoogleMap.js
```diff
@@ -242,6 +243,7 @@ ReactiveGoogleMap.defaultProps = {
 	showMarkerClusters: true,
 	unit: 'mi',
 	defaultRadius: 100,
+	mapService: 'GoogleMap',
```
lakhansamani, line 246: This prop is internal to `ReactiveMap` and shouldn't be exposed to end-user. There are more chances of user misusing it. example

`<ReactiveGoogleMap mapService="OpenStreentMap" ... />` and this will break the maps.

lakhansamani (approved):

lakhansamani:
@e-gaulue thank you for this contribution 👍 

I resolved the above + made some changes with `getLocationObject` function as it was breaking for `searchAsMove` on OpenStreet map

Here is the commit: https://github.com/appbaseio/reactivesearch/commit/8c9d0747f9f0a45ff43961a4ef1b0c5f5d952eb4

PR #1366, opened 2020-02-17
feat(reactivemap): allow GeoDistance tools without Google Map

e-gaulue:
GeoDistance tools doesn't work without GoogleMap. Examples shown on the Internet give the feeling it's alright because google map is already present in the page. But when using OpenStreetMap only, you may want to totally remove the need for Google Map.

Now look if Google Map is loaded before calling its functions. Avoid undefined error.

Add a `geocoder` prop to define your own geocoder. Should be an object with at least a `geocode` function which mimic Google's one. If tou have positioned data in your elastic, you could for instance just use their position string. Example:
```
geocoder={{geocode: function (coordinatesObj, todo) {
       const [tlat,tlng] = coordinatesObj.address.split(',');
             todo([{geometry: {location: {lat: () => tlat, lng: () => tlng}}}],'OK');
       }
}}
```

### DIFF
--- a/packages/maps/src/components/basic/GeoCode.js
+++ b/packages/maps/src/components/basic/GeoCode.js
@@ -8,37 +8,45 @@ class GeoCode extends Component {
 				lng: location.coords.longitude,
 			};
 
-			this.geocoder.geocode({ location: coordinatesObj }, (results, status) => {
-				if (status === 'OK') {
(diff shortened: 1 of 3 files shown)


lakhansamani (approved):

lakhansamani:
@e-gaulue thank you for the contribution 👍 💯

PR #1369, opened 2020-02-26
fix(date range picker): fix controlled date range picker

mitchgillin:

- [x] Describe the proposed changes and how it'll improve the library experience.
Fixes the DateRange component so that it keeps tracking state properly when it is a controlled component. 

- [x] Please make sure that there is no linting errors in the code.

- [x] Create a PR to add/update the docs (if needed) at [here](https://github.com/appbaseio/Docs).
NA: this change will follow the docs. 

- [x] Create a PR to add/update the storybook (if needed) at [here](https://github.com/appbaseio/playground).
NA


mitchgillin:
Need to update with an additional fix

PR #1418, opened 2020-04-17
feat(web): reset query for ReactiveComponent on clicking selectedfilter

### LINKED ISSUE #1414: Selected Filters (Clear All) doesn't work with ReactiveComponent
**Affected Projects**
- [x] ReactJS
- might be VueJS (not sure)

**Library Version:** x.y.z
- 3.7.11

**To Reproduce**
- Click any value from the cars list. Then select `Clear All` button.

**Expected behavior**
- The filter should get reset and all data should be available.

**CSB**
https://codesandbox.io/s/reactivecomponent-ffipt?file=/src/index.js

jyash97:
Screen record: https://www.loom.com/share/e3be3f8cc7744b198554fccd9aafdb9a

Fixes #1414

### DIFF
--- a/packages/web/src/components/basic/ReactiveComponent.js
+++ b/packages/web/src/components/basic/ReactiveComponent.js
@@ -159,6 +159,19 @@ class ReactiveComponent extends Component {
 			);
 		}
 
+		checkPropChange(this.props.selectedValue, prevProps.selectedValue, () => {
+			/*
+				Reset query when SelectedFilters are clicked. Note: `selectedValue` becomes null.
+			*/
+
+			if (this.props.selectedValue === null) {
+				this.props.updateQuery({
+					componentId: this.props.componentId,
+					query: null,
+				});
+			}
+		});
+
 		if (this.props.defaultQuery && !isEqual(this.props.defaultQuery(), this.defaultQuery)) {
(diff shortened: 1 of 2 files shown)


packages/web/src/components/basic/ReactiveComponent.js
```diff
@@ -159,6 +159,19 @@ class ReactiveComponent extends Component {
 			);
 		}
 
+		checkPropChange(this.props.selectedValue, prevProps.selectedValue, () => {
+			/*
+				Reset query when SelectedFilters are clicked. Note: `selectedValue` becomes null.
+			*/
+
+			if (this.props.selectedValue === null) {
+				this.props.updateQuery({
```
bietkul, line 168: @jyash97 Instead of updating the query you should set the value to `null`. Use `setValue` method.
This code will not work when `enableAppbase` is `true`.

bietkul (changes requested):

packages/web/src/components/basic/ReactiveComponent.js
jyash97: I have tested with enableAppbase as well it works.

jyash97: Do check the video attached

PR #1423, opened 2020-04-20
fixes korean typing 

littlehome-eugene:

screenshot was from `dejavu`

Before fix 

broken

after fix

fixed

- [ x] Describe the proposed changes and how it'll improve the library experience.
  as shown with the gif, currently, it's hard to type korean (i guess other languages too which requires several keystroke to make up a character)
The pull requests fixes it.
- [ x] Please make sure that there is no linting errors in the code.
I just deleted a line which I believe has no effect on other functionality, there could be no linting errors
- [x ] Create a PR to add/update the docs (if needed) at [here](https://github.com/appbaseio/Docs).
no docs required
- [x ] Create a PR to add/update the storybook (if needed) at [here](https://github.com/appbaseio/playground).

let me know if I need to fix any

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -35,7 +35,7 @@
 		"@appbaseio/reactivecore": "9.5.5",
 		"appbase-js": "4.0.2",
 		"cross-env": "^5.2.0",
-		"downshift": "^5.1.1",
+		"downshift": "^1.31.2",
 		"emotion": "^9.0.0",
 		"emotion-theming": "^9.0.0",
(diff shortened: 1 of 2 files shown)


littlehome-eugene:
let me know what `label` fails means :(

bietkul (changes requested):
@littlehome-eugene Thanks for the PR but this change will not work because you're changing the `input` component to the `un-controlled` component but we're using it as a `controlled` component which means that the `value` prop is essential.

PR #1427, opened 2020-04-21
fix(web): voice search stops too soon

ShahAnuj2610:
the mic was not able to detect sentences with multiple words. It was just taking first word and processing. This PR fixes the same.

Check the video with this fix: https://www.loom.com/share/6bd5ce6e211f4403aefec9036e9c72f8

Read more about interim results over here: https://wicg.github.io/speech-api/#dom-speechrecognition-interimresults

### DIFF
--- a/packages/web/src/components/search/addons/Mic.js
+++ b/packages/web/src/components/search/addons/Mic.js
@@ -74,7 +74,7 @@ class Mic extends React.Component {
 			}
 			this.instance = new SpeechRecognition();
 			this.instance.continuous = true;
-			this.instance.interimResults = true;
+			this.instance.interimResults = false;
 			this.instance.lang = lang;
 			if (getInstance) {
 				getInstance(this.instance);


packages/web/src/components/search/addons/Mic.js
```diff
@@ -74,7 +74,7 @@ class Mic extends React.Component {
 			}
 			this.instance = new SpeechRecognition();
 			this.instance.continuous = true;
-			this.instance.interimResults = true;
+			this.instance.interimResults = false;
```
jyash97, line 77: Cool, this looks like the issue, were we stopping the mic as soon as we got the result? I think we forgot to add the check for whether the response received is final or not.

jyash97 (approved):
LGTM 💯

packages/web/src/components/search/addons/Mic.js
ShahAnuj2610: basically `onresult` gets called on each word results in a sentence. And we were setting values in the state only for first word and stopping the mic. But now `onresult` won't get called unless and until we stop to speak.

jyash97: Cool makes sense 👍

PR #1428, opened 2020-04-21
fix(web): make size and sortBy prop reactive for MultiList

### LINKED ISSUE #1415: sortBy is not reactive in MultiList
**Affected Projects**
- [x] React
might be **Vue** also

**Describe the bug**
- on changing `sortBy` the list doesn't get updated

**To Reproduce**
- https://opensource.appbase.io/playground/?path=/story/list-components-multilist--custom-sort

**Expected behavior**
- the list should get updated according to the sort value passed

**Video**
https://www.loom.com/share/4defcbe892db45f08b87bac3ce86a852

ShahAnuj2610:
fix #1415 

## Demo video
https://www.loom.com/share/5286b0a81f304e4a957dee7095b9a82e

### DIFF
--- a/packages/web/src/components/list/MultiList.js
+++ b/packages/web/src/components/list/MultiList.js
@@ -170,6 +170,10 @@ class MultiList extends Component {
 			this.updateQuery(Object.keys(this.state.currentValue), this.props);
 		}
 
+		checkSomePropChange(this.props, prevProps, ['size', 'sortBy'], () =>
+			this.updateQueryOptions(this.props),
+		);
+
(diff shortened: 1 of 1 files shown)


bietkul (changes requested):
@ShahAnuj2610 Make that change in all list components.
And don't make `size` as reactive.

ShahAnuj2610:
> @ShahAnuj2610 Make that change in all list components.
> And don't make `size` as reactive.

@bietkul it's reactive for other list components. Also size prop is reactive for the list components.

bietkul (approved):

PR #1442, opened 2020-05-03
style(web): normalize end of line rule across platform

nikhil-varma:
The current End of line rules does not go well with Windows and eslint shows ton of errors. This deteriorate the developer experience. VSCode auto converts line ending to CLRF which does not adhere to the eslint rules. A quick fix from Microsoft is in works but enforcing in project-wide would be much more cleaner and independent of the IDE.

- [x] Describe the proposed changes and how it'll improve the library experience.
- [x] Please make sure that there are no linting errors in the code.
- [ ] Add a demo video/gif/screenshot to explain how did you test the fix.
- [ ] If it is a global change, try to add any side effects that it could have.
- [ ] Create a PR to add/update the docs (if needed) at [here](https://github.com/appbaseio/Docs).
- [ ] Create a PR to add/update the storybook (if needed) at [here](https://github.com/appbaseio/playground).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,6 +19,10 @@
 		"no-tabs": 0,
 		"no-underscore-dangle": 0,
 		"operator-linebreak": ["error", "before"],
+		"linebreak-style": [
+			"error",
+			"unix"
+		  ], 
 		"prefer-destructuring": 0,
 		"no-console": [
 			"error",
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+* text=auto
+
+*.js text eol=lf
+
+# The test fixtures are text files.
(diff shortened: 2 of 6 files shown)


nikhil-varma:
@jyash97 Should we add these rules for all sub-projects (vue, web, etc...)? Also, is anyone else facing such issues? If so I think this PR would help create consistency across platforms.. I made this work for web as of now. Can do the same for Vue and other projects as well.

nikhil-varma:
@siddharthlatest I think we can review this and close it.

siddharthlatest (approved):
This lgtm (y)

Will merge this in the next release. Thanks @nikhil-varma

bietkul (approved):

PR #1443, opened 2020-05-05
fix(web): hide mic for browsers not supporting SpeechRecognition

ShahAnuj2610:
## What does this PR do?
- hide mic if browser does not support SpeechRecognition

### DIFF
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -755,7 +755,7 @@ class CategorySearch extends Component {
 						{this.renderCancelIcon()}
 					</InputIcon>
 				)}
-				{showVoiceSearch && (
+				{this.shouldMicRender(showVoiceSearch) && (
 					<Mic
 						getInstance={getMicInstance}
 						render={renderMic}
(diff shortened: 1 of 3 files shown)


jyash97:
@ShahAnuj2610 I think we decided to not show the mic when the browser is not supported. cc @siddharthlatest  can you confirm ?

ShahAnuj2610:
@jyash97 made the change

bietkul (approved):

packages/web/src/components/search/CategorySearch.js
```diff
@@ -777,6 +777,12 @@ class CategorySearch extends Component {
 		);
 	};
 
+	shouldMicRender(showVoiceSearch) {
+		// checks for SSR
+		if (typeof window === 'undefined') return false;
```
jyash97, line 782: Quick question @ShahAnuj2610  

Are we not rendering Mic in SSR ?

ShahAnuj2610: @jyash97 we are. In case of SSR on initial render, the window variable is undefined, so this is just a defensive check for SSR.

jyash97: Cool, yeah just wanted to clarify that we are showing it back or not after didMount.

PR #1448, opened 2020-05-08
fix(vue): dataField optional for DataSearch for rs-api

ShahAnuj2610:
## Pre-requisite
Read this https://stackoverflow.com/questions/56904327/access-props-value-in-another-props-validator

## What does this PR do?
- fix validation for dataField when `enableAppbse=true`

## Test
https://www.loom.com/share/4f6b6af7973441dfb53c758ce8d02623

## Docs
https://github.com/appbaseio/Docs/pull/141


packages/vue/src/components/search/DataSearch.jsx
```diff
@@ -262,15 +264,31 @@ const DataSearch = {
 		},
 	},
 	methods: {
+		validateDataField() {
+			const propName = 'dataField';
+			const componentName = 'DataSearch';
```
bietkul: @ShahAnuj2610 Use it from `name` property.
https://github.com/appbaseio/reactivesearch/blob/bb371f05df1dce890f0b1d6bac6a6288922defae/packages/vue/src/components/search/DataSearch.jsx#L54

packages/vue/src/components/search/DataSearch.jsx
```diff
@@ -262,15 +264,31 @@ const DataSearch = {
...
 		updateQueryOptions() {
-			if (
-				this.customHighlight
-				&& typeof this.customHighlight === 'function'
-			) {
-				this.setCustomHighlightOptions(
-					this.componentId,
-					this.customHighlight(this.$props),
-				);
+			if (this.customHighlight && typeof this.customHighlight === 'function') {
```
bietkul, line 304: @ShahAnuj2610 Why formatting changes?

bietkul (changes requested):

packages/vue/src/components/search/DataSearch.jsx
ShahAnuj2610: not sure. I'm using prettier and eslint as per repo config.

packages/vue/src/components/search/DataSearch.jsx
ShahAnuj2610: done

bietkul: @ShahAnuj2610 How is it done?
You're still hardcoding the `componentName`, it should be `this.name`.

bietkul (changes requested):

packages/vue/src/components/search/DataSearch.jsx
ShahAnuj2610: @bietkul check this https://github.com/appbaseio/reactivesearch/pull/1448/files#diff-29c2691a20b54ace57eca812d2d9ca86R269

bietkul (approved):

PR #1466, opened 2020-05-18
refactor(vue): watch for SingleList.jsx

ShahAnuj2610:
## What this PR does?
Fix props getting updated in Redux store

## Test
https://www.loom.com/share/20eabb8891fa4abf8543494d43e650ac

### DIFF
--- a/packages/vue/src/components/list/SingleList.jsx
+++ b/packages/vue/src/components/list/SingleList.jsx
@@ -101,17 +101,6 @@ const SingleList = {
 	},
 	mounted() {
 		this.setReact(this.$props);
-		const propsKeys = getValidPropsKeys(this.$props);
-		this.$watch(propsKeys.join('.'), (newVal, oldVal) => {
-			checkSomePropChange(newVal, oldVal, propsKeys, () => {
-				this.updateComponentProps(this.componentId, this.$props, componentTypes.singleList);
-				this.updateComponentProps(
-					this.internalComponent,
-					this.$props,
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/list/SingleList.jsx
```diff
@@ -132,6 +121,17 @@ const SingleList = {
 		this.removeComponent(this.internalComponent);
 	},
 	watch: {
+		$props: {
+			deep: true,
+			handler(newVal) {
+				this.updateComponentProps(this.componentId, newVal, componentTypes.singleList);
```
bietkul: @ShahAnuj2610 What is `newVal` here?
Can you log it and share a video?
Also for the performance reasons only update the valid props.

bietkul (changes requested):

packages/vue/src/components/list/SingleList.jsx
ShahAnuj2610: @bietkul 
> What is newVal here?

it's the new value for props. Will share the video.
> Also for the performance reasons only update the valid props

it's already handled in `updateComponentProps`
https://github.com/appbaseio/reactivecore/blob/c6deb6e40da8da8eb2ab7b1397ce838098c61e58/src/actions/props.js#L35
https://github.com/appbaseio/reactivecore/blob/c6deb6e40da8da8eb2ab7b1397ce838098c61e58/src/actions/props.js#L4

PR #1482, opened 2020-05-28
fix(vue): logic to build default SingleList filter query #1481

### LINKED ISSUE #1481: SingleList default query for missing filter column value gets overwritten
**Affected Projects**
Vue.JS

**Library Version:** 
1.4.0

**Describe the bug**
showing items with missing values displays a filter option for it, and it displays the doc_count correctly.

However, when we select such option, no results are returned, it should return some since the doc_count is greater than 0.

**To Reproduce**
1. You need an index with an aggregation column and items missing that column.
1. add a SingleList component with property `  :showMissing="true"`
     - the N/A option is displayed and shows the doc_count
     - when selecting the N/A option the ReactiveList component shows no results

**Expected behavior**
1. selecting the N/A option should display some results.

**Additional context**
* Looking at the code we can see the Default Query handles this situation properly: [SingleList.jsx L481](https://github.com/appbaseio/reactivesearch/blob/5fc8a847f7c1b91fa53f9584cb4e34e6af3b47e6/packages/vue/src/components/list/SingleList.jsx#L481)
* However, right after, the query is overwritten with the query that would be used for a specific term: [SingleList.jsx L490](https://github.com/appbaseio/reactivesearch/blob/5fc8a847f7c1b91fa53f9584cb4e34e6af3b47e6/packages/vue/src/components/list/SingleList.jsx#L490)
* Reversing the order of these 2 blocks fixes the issue.

acosta-edgar:
Fixes https://github.com/appbaseio/reactivesearch/issues/1481
The query that handles the case when the user selects to find items with no value for the filter field gets overwritten by the query that uses the selected term value.

The solution is to create the term query by default, and overwrite it with the missing value query if necessary.

### DIFF
--- a/packages/vue/src/components/list/SingleList.jsx
+++ b/packages/vue/src/components/list/SingleList.jsx
@@ -478,6 +478,11 @@ SingleList.defaultQuery = (value, props) => {
 		};
 	}
 	if (value) {
+		query = {
+			term: {
+				[props.dataField]: value,
+			},
+		};
 		if (props.showMissing && props.missingLabel === value) {
 			query = {
 				bool: {
@@ -487,11 +492,6 @@ SingleList.defaultQuery = (value, props) => {
 				},
 			};
 		}
-		query = {
-			term: {
-				[props.dataField]: value,
-			},
-		};
 	}
 
 	if (query && props.nestedField) {


acosta-edgar:
@aol-nnov @gaving @literarymachine @willopez any update or feedback on this PR?
Thanks!

acosta-edgar:
@siddharthlatest ?

acosta-edgar:
@jyash97 is there any way to get attention to this PR?
it fixes a bug that only appears on very specific circumstances.
However, the fix is straightforward.
It should be easy to check and move forward.

jyash97:
Hey @acosta-edgar Thanks for the PR 😄 

I will check the changes over this weekend. If you can share a codesandbox link of the issue it would help me a lot in the process of reviewing.

acosta-edgar:
Hi @jyash97 I can try to figure out how to do the codesandbox thing. However, it heavily depends on both the query and the data. I don't know if I can set the sample data for this.

acosta-edgar:
Hi @jyash97 , here is the codesandbox that reproduces the bug: https://codesandbox.io/s/singlelist-oz2w5

It is a fork of https://codesandbox.io/s/github/appbaseio/reactivesearch/tree/dev/packages/web/examples/SingleList

The only change is on file `src/index.js`, at line 26 I added:

``` 
              showMissing={true}
              missingLabel="NONE"
```
Which has the effect of adding a filter option "NONE" at the Series filter (SingleList).
This works fine: you will see the option "NONE" with 21 hits.

The bug is that when you select the "NONE" option you get no results, but it should return 21 books!
That is the issue my PR fixes.

Thanks!

bietkul (approved):

bietkul:
@acosta-edgar Thanks for the PR and nice explanation. It'll be released in the next version.

acosta-edgar:
Awesome! Thanks @bietkul !

PR #1489, opened 2020-06-08
chore(version-scripts): add deploy storybook for maps & ignore next file

jyash97:
1. Add Storybook Deploy in ReactiveMaps post publish
2. Ignore .next folder generated in ssr apps

### DIFF
--- a/scripts/version-update.sh
+++ b/scripts/version-update.sh
@@ -32,7 +32,7 @@ var walk = function(dir,done) {
 
 walk("$var", function(err, results) {
     if (err) throw err;
-    var filesToUpdate = results.filter(item => item.includes("package.json")).filter(item => !item.includes("node_modules"));
+    var filesToUpdate = results.filter(item => item.includes("package.json")).filter(item => !item.includes("node_modules") && !item.includes(".next"));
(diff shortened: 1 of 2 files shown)


scripts/version-update.sh
```diff
@@ -32,7 +32,7 @@ var walk = function(dir,done) {
 
 walk("$var", function(err, results) {
     if (err) throw err;
-    var filesToUpdate = results.filter(item => item.includes("package.json")).filter(item => !item.includes("node_modules"));
+    var filesToUpdate = results.filter(item => item.includes("package.json")).filter(item => !item.includes("node_modules") && !item.includes(".next"));
```
bietkul: @jyash97 Instead of just ignoring `.next` folder can you ignore all folders that start with a dot?

bietkul (changes requested):

scripts/version-update.sh
jyash97: Fixed. Fiddle to see the logic https://jsfiddle.net/zjv0ue7L/

bietkul (approved):

bietkul (approved):

PR #1505, opened 2020-07-03
feat(vue): add scrollOnChange option to prevent scrolling to top

jessevanmuijden:
When users change the filter options and fewer hits are found than before changing the filter, the page jumps to the top. It should be possible to disable this automatic scrolling behavior. This has been implemented in the React version of the library some time ago (see https://github.com/appbaseio/reactivesearch/issues/528).

### DIFF
--- a/packages/vue/src/components/result/ReactiveList.jsx
+++ b/packages/vue/src/components/result/ReactiveList.jsx
@@ -114,7 +114,7 @@ const ReactiveList = {
 		pagination: VueTypes.bool.def(false),
 		paginationAt: types.paginationAt.def('bottom'),
 		react: types.react,
-		scrollOnChange: VueTypes.bool.def(false),
+		scrollOnChange: VueTypes.bool.def(true),
 		showResultStats: VueTypes.bool.def(true),
 		showEndPage: VueTypes.bool.def(false),
 		size: VueTypes.number.def(10),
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/result/ReactiveList.jsx
```diff
@@ -114,6 +114,7 @@ const ReactiveList = {
 		pagination: VueTypes.bool.def(false),
 		paginationAt: types.paginationAt.def('bottom'),
 		react: types.react,
+		scrollOnChange: VueTypes.bool.def(false),
```
jyash97: Hey thanks for the PR 💯 

default value should be `true` like it is in React version

jyash97 (approved):
Thanks @jessevanmuijden 

Looks good to me ✨  Waiting for another approval & we can merge it than.

jyash97:
@jessevanmuijden  You have missed this one place -> https://github.com/appbaseio/reactivesearch/blob/82c08cbdce0653a376dba5507313f8edefa654ae/packages/vue/src/components/result/ReactiveList.jsx#L269

You may have to do the changes here as well.

bietkul (approved):

bietkul:
@jessevanmuijden Thanks for the PR, it'll be out in next release.

PR #1517, opened 2020-07-31
feat(web): add support for include & exclude fields in Search Components

jyash97:
Add include and exclude Fields support in DataSearch component. 

### Testing DataSearch

**_msearch**

https://www.loom.com/share/3ad99493993e4f369065e498eed25048

**RS API**
https://www.loom.com/share/e69bd070f89c47da82daa168c3a83e61

### Testing CategorySearch

https://www.loom.com/share/9b44d372b2e34d32a60977b762f60f86


packages/web/src/components/search/DataSearch.js
```diff
@@ -1064,7 +1067,9 @@ DataSearch.defaultProps = {
 	downShiftProps: {},
 	enableSynonyms: true,
 	enableQuerySuggestions: false,
+	excludeFields: [],
 	iconPosition: 'left',
+	includeFields: ['*'],
```
bietkul, line 1072: @jyash97 Remove the default values.

bietkul (changes requested):
@jyash97 
1. Remove the default values.
2. Does source filtering affects the query for other components that has a `react` dependency on `DataSearch`? Ideally it shouldn't it should only affect the suggestions query but as per your logic I think it will change. Please share a video for the same.

packages/web/src/components/search/DataSearch.js
jyash97: Do we need it as a required prop ?

jyash97:
> Does source filtering affects the query for other components that has a react dependency on DataSearch? Ideally it shouldn't it should only affect the suggestions query but as per your logic I think it will change. Please share a video for the same.


https://www.loom.com/share/77d6049cbefe49e6a144e4200f85f201

PR #1527, opened 2020-08-11
Fix/vue/use default sort option when updating sort

acosta-edgar:
# Problem

`ReactiveList` executes method `updateQueryOptions()` when some properties get modified. However, it ignores the `defaultSortOption` property when it is set.

## Code Sandbox

In [this code sandbox](https://codesandbox.io/s/wispy-night-o1rj6) we added properties `sortOptions` and `defaultSortOption`.

1. Notice, that when loading the page:
    * it shows the Sort selector, which has 3 options:
        * Title (descending)
        * Relevance (descending)
        * Rating (average rating descending)
    * it shows Rating as the default sort order
        * as specified by property `defaultSortOption` passed to component ReactiveList.
    * items are sorted by average rating accordingly
1. Now when we submit a search term:
    * this code modifies property `dataField` of the `ReactiveList` component, via `data.sort`
    * in turn, this triggers the execution of method `ReactiveList.updateQueryOptions()`

**Note:** the search is executed however
* the results are sorted by Title (descending)
* whereas the selected Sort option is still the default (Rating)

This is because `ReactiveList.updateQueryOptions()` is always taking the first sort options, which is not necessarily the default one.

# Solution
We find and use the default Sort Option to set the sort Query option.

# Check list

- [*] Describe the proposed changes and how it'll improve the library experience.
- [*] Please make sure that there are no linting errors in the code.
- [*] Add a demo video/gif/screenshot to explain how did you test the fix.
- [ ] If it is a global change, try to add any side effects that it could have.
- [ ] Create a PR to add/update the docs (if needed) at [here](https://github.com/appbaseio/Docs).
- [ ] Create a PR to add/update the storybook (if needed) at [here](https://github.com/appbaseio/playground).


acosta-edgar:
something does not look good with this PR, I'll try to create it again.

PR #1584, opened 2021-01-06
Changed the event names from camelCase to kebab-case

royduin:
As of: https://vuejs.org/v2/guide/components-custom-events.html#Event-Names, event names should be kebab-case. Got issues with this with DOM templates.

### DIFF
--- a/packages/vue/src/components/basic/DownShift.jsx
+++ b/packages/vue/src/components/basic/DownShift.jsx
@@ -358,10 +358,11 @@ export default {
 			Object.keys(stateToSet).map(key => {
 				// eslint-disable-next-line
 				this.isControlledProp(key)
-					? this.$emit(`${key}-change`, stateToSet[key])
+					? this.$emit(`${key}Change`, stateToSet[key]) && this.$emit(`${key}-change`, stateToSet[key])
 					: (this[`internal_${key}`] = stateToSet[key]);
 			});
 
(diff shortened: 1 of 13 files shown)


bietkul (approved):
lgtm :+1:

bietkul (changes requested):
@royduin This PR would break the events for existing users.
We should make that change with backward compatibility by emitting the events for both cases (kebab-case and camelCase). 
In v2.0 we'll remove the camelCase events.

royduin:
I've re-added them 👍🏻

packages/vue/src/components/basic/DownShift.jsx
```diff
@@ -358,11 +358,12 @@ export default {
 			Object.keys(stateToSet).map(key => {
 				// eslint-disable-next-line
 				this.isControlledProp(key)
-					? this.$emit(`${key}Change`, stateToSet[key])
+					? this.$emit(`${key}Change`, stateToSet[key]) && this.$emit(`${key}-change`, stateToSet[key])
```
bietkul, line 361: @royduin Can we remove `&&` operator and use `if else` condition to emit events?

bietkul (changes requested):
@royduin A minor change is needed. The rest of the changes look good to me.

bietkul:
I'm merging it so it can be out in today's version.

royduin:
Thanks you! 💯

PR #1597, opened 2021-01-22
refactor(vue index file): remove unused import

larsdouweschuitema:
**Issue type:** Refactor
**Platform:** Vue
**Description:** Clean up linter warning by removing unused import. That is not only unused, but also commented out using ESLint, to have a cleaner console when running the Vue application.

**Linter warning**
```
[WEB] [build.cjsWatch] (!) Unused external imports
[WEB] [build.cjsWatch] polyfills imported from external module '@appbaseio/reactivecore' but never used
```

**Before console**
image

**After console**
image

**Loom video**
Video shows you that the console is clean when removing the import. And also shows you that the application is still working when removing that specific line, and no errors are shown.
https://www.loom.com/share/660779d47ad946c986877b4376cfa57a

### DIFF
--- a/packages/vue/src/index.js
+++ b/packages/vue/src/index.js
@@ -1,7 +1,3 @@
-// Add polyfills to support in IE
-// eslint-disable-next-line
-import { polyfills } from '@appbaseio/reactivecore';
-
 import ReactiveList from './components/result/ReactiveList.jsx';
 import ReactiveBase from './components/ReactiveBase/index.jsx';
 import DataSearch from './components/search/DataSearch.jsx';


bietkul:
@larsdouweschuitema Thanks for the PR. FYI, we're importing this file intentionally to support polyfills for browsers that don't support some functionalities. Please check at here https://github.com/appbaseio/reactivecore/blob/master/src/utils/polyfills.js

I think we can make that change in `reactivecore` to wrap polyfills in a function and then call it in the root file of vue.
Wdyt?

larsdouweschuitema:
> @larsdouweschuitema Thanks for the PR. FYI, we're importing this file intentionally to support polyfills for browsers that don't support some functionalities. Please check at here https://github.com/appbaseio/reactivecore/blob/master/src/utils/polyfills.js
> 
> I think we can make that change in `reactivecore` to wrap polyfills in a function and then call it in the root file of vue.
> Wdyt?

Ok I understand what you are doing with the unused import. About your idea: I agree. That way we have a clean console and can still import the polyfills.

Issue #1526, opened 2020-08-10
isEqual does not properly check date objects


mitchgillin:
**Affected Projects**
React

**Library Version:** x.y.z
9.6.15

**Describe the bug**
In the date range component, the `isEqual` function is passed an object with an object as the value in the key value pair 
`const d1 = new Date("1-15-2020);`
`const d2 = newDate("2-15-2020);`
`const d3 = new Date("3-15-2020);`
`isEqual({start: d1, end: d2}, {start:d1, end:d3} === true`

**To Reproduce**
https://codesandbox.io/s/stoic-bush-jsrrt?file=/src/App.js:164-184
   Use a date range as a controlled component; the on change methods will not fire because it is detecting the values are equal. 

**Expected behavior**
Clicking the button in the sandbox above should set the date properly; this can be fixed by adding a JSON.stringify check to the isEqual function

**Screenshots**

**Desktop (please complete the following information):**
 - OS:  MacOS 10.15.6 
 - Browser: Brave
 - Version: 1.13.59

**Additional context**

PR #1600, opened 2021-01-28
fix(web): isEqual does not properly check date objects

anik-ghosh-au7:
#1526 
Issue Type:

- Bug

Platform:

- Web

Description:

added JSON.stringify check to the isEqual function inside the dateRange Component, when the component updates

https://www.loom.com/share/b3f7e0df39d944898a14fe0219f85314

### DIFF
--- a/packages/web/src/components/date/DateRange.js
+++ b/packages/web/src/components/date/DateRange.js
@@ -61,7 +61,7 @@ class DateRange extends Component {
 	}
 
 	componentDidUpdate(prevProps) {
-		if (!isEqual(this.props.value, prevProps.value)) {
+		if (!isEqual(JSON.stringify(this.props.value), JSON.stringify(prevProps.value))) {
 			this.handleDateChange(this.props.value, false, this.props);
 		} else {
 			const { currentDate } = this.state;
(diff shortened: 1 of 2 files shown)


packages/web/src/components/result/ReactiveList.js
```diff
@@ -661,6 +661,7 @@ class ReactiveList extends Component {
 			css={sortOptions}
 			className={getClassName(this.props.innerClass, 'sortOptions')}
 			name="sort-options"
+			aria-label="Sort options"
```
bietkul: @anik-ghosh-au7 Why do we have this change here?

packages/web/src/components/date/DateRange.js
```diff
@@ -61,7 +61,7 @@ class DateRange extends Component {
 	}
 
 	componentDidUpdate(prevProps) {
-		if (!isEqual(this.props.value, prevProps.value)) {
+		if (!isEqual(JSON.stringify(this.props.value), JSON.stringify(prevProps.value))) {
```
bietkul, line 64: @anik-ghosh-au7 I'm not sure if this change would fix the issue because `isEqual` does the same thing. It'll be best if you can share the kind of comparison that fails.
https://www.loom.com/share/0105a5a6b3a7490faaf696edfeb1f0fc

bietkul (changes requested):

packages/web/src/components/date/DateRange.js
anik-ghosh-au7: @bietkul https://www.loom.com/share/84cceda1046243e982655727eed14d3c

bietkul (approved):

PR #1602, opened 2021-02-03
fix(vue): array values from URL parameters

VincentBean:
When using a MultiList, URLParams and an array in the URL it will convert the array to a string
For example:
["Jacket", "Hooded"] will be converted to ["Jacket,Hooded"]


bietkul (approved):
Tested it. Looks good 👍

PR #1607, opened 2021-02-07
Prevent showLoadMore from causing setState error

rachaeldawn:
- Mimic the setValue signature by adding a `hasMounted = true` as a parameter at the end of updateQueryOptions
- When checking `props.showLoadMore`, respond to whether or not the component has mounted by re-assigning state
- Update the constructor to use the new parameter for updateQueryOptions

Result: using `showLoadMore` on MultiList should no longer throw `setState` errors.

This shouldn't have linting errors, as I have mimicked all of the code style to match the file, but I can't get the lint script to run without errors flooding my console that do not pertain to the code I have changed.


bietkul (approved):
lgtm :+1:

PR #1616, opened 2021-02-18
Patch 3

rachaeldawn:
Okay, so I found a really edge-case bug that was causing an extremely frustrating user experience in our app (OnTopical).

If `componentDidUpdate` gets called, and `value` is set, the call to the store to execute the search query is called regardless of it being the user's responsibility. In our experience, this only happens the first time that the DataSearch component is loaded, and never again.

Call stack issues are:
`this.setValue` leads to line 423, where `this.updateQuery` is called
`this.updateQuery` erroneously dispatches a `props.setQueryOptions` without passing the `execute = false` argument (third argument, [see this](https://github.com/appbaseio/reactivecore/blob/master/src/actions/query.js#L510)).

I am not certain there are no linting errors, but I have done what I can to ensure code-style consistency. To test the fix, I had to manually change the code in my node_modules, slowly following along to find what happened. I noticed that every single time the `setQueryOptions` connection to the props gets called, it _also_ executed the query when the expected behavior is not to.

The storybook / docs are unnecessary, as this amendment simply guarantees what the docs assert.

### DIFF
--- a/packages/web/src/components/search/CategorySearch.js
+++ b/packages/web/src/components/search/CategorySearch.js
@@ -568,13 +568,14 @@ class CategorySearch extends Component {
 			updateCustomQuery(props.componentId, props, value);
 		}
 
-		// query options should be applied to the source component,
-		// not on internal component, hence using `this.props.componentId` here
-		props.setQueryOptions(props.componentId, {
-			...this.queryOptions,
-			...customQueryOptions,
-		});
 		if (!this.isPending) {
+			// execute the query on an uncontrolled component
+      // query options should be applied to the source component,
+      // not on internal component, hence using `this.props.componentId` here
+      props.setQueryOptions(props.componentId, {
+        ...this.queryOptions,
+        ...customQueryOptions,
+      });
 			props.updateQuery({
 				componentId: props.componentId,
 				query,
--- a/packages/web/src/components/search/DataSearch.js
+++ b/packages/web/src/components/search/DataSearch.js
@@ -510,10 +510,20 @@ class DataSearch extends Component {
 
 		// query options should be applied to the source component,
 		// not on internal component, hence using `this.props.componentId` here
-		props.setQueryOptions(props.componentId, {
-			...this.queryOptions,
-			...customQueryOptions,
-		});
+		if ('value' in props) {
+			// do NOT execute a query if the component is controlled. That's the user's responsibility.
+			props.setQueryOptionsNoExecute(props.componentId, {
+				...this.queryOptions,
+				...customQueryOptions,
+			});
+		} else {
+			// execute the query on an uncontrolled component
+			props.setQueryOptions(props.componentId, {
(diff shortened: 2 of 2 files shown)


bietkul (changes requested):
@rachaeldawn Thanks for the PR.
There is a property called `isPending` that controls the controlled behavior in a way that when it is `false` then only the `DataSearch` would trigger query.
I think you can move the following code https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/search/DataSearch.js#L513-#L516 to https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/search/DataSearch.js#L517-#L527 block.
Please test it once and lmk if that works.
You can make that change in `CategorySearch` too.

rachaeldawn:
> @rachaeldawn Thanks for the PR.
> There is a property called `isPending` that controls the controlled behavior in a way that when it is `false` then only the `DataSearch` would trigger query.
> I think you can move the following code https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/search/DataSearch.js#L513-#L516 to https://github.com/appbaseio/reactivesearch/blob/next/packages/web/src/components/search/DataSearch.js#L517-#L527 block.
> Please test it once and lmk if that works.
> You can make that change in `CategorySearch` too.

I am very happy to report that it did indeed fix the problem for me. I will amend this commit and tag you once that has been changed.

rachaeldawn:
@bietkul 
In my testing with the data search on my machine, this worked for me. Applied the same change to the category search, and made sure to keep the comments so future readers understand why that code is there.

bietkul (approved):
lgtm 👍

rachaeldawn:
@bietkul 
By the way, I just want to express my gratitude for jumping on this so fast. I know it's an absolute hassle having to deal with extremely un-helpful Issues, so I figured a PR would be better. I was anticipating at least a couple days, but you got back to me really really fast and I appreciate that a lot! 

I hope your day is going well! :D

bietkul:
@rachaeldawn A new version(`v3.15.0`) is out with this fix.

rachaeldawn:
> @rachaeldawn A new version(`v3.15.0`) is out with this fix.

I just spotted that, thank you!

Is there by chance a Slack channel, or a Discord server I can join to ask questions about the architecture / inner workings of the ReactiveSearch library? This is my third PR, and I get the sense it's going to be less effort for me to just be an active (but informed) contributor than it is going to be to have to roll my own components in the future. My main goal is to be able to ask questions like "If I make this change, what does it affect? When I call this method, does it impact <x>" and stuff like that, but GitHub issues doesn't really seem like the appropriate place to ask those sorts of questions.

bietkul:
@rachaeldawn  We don't have a discord/slack channel as of now. Currently, all communication happens on Github and we try to answer straightforward questions as soon as possible. You can try Github discussions to ask questions.

Issue #1615, opened 2021-02-17
clearValues() function of SelectedFilters component removes all query parameters.


divyadeep-MT:
**Affected Projects**
React

**Library Version:** x.y.z
-

**Describe the bug**
`clearValues()` function of `SelectedFilters` component removes all query parameters, including those that aren't a part of ReactiveSearch's filters/components.

**To Reproduce**
Simply add an additional query parameter along with a few filters, and use the `SelectedFilter` component's `clearValues()` function inside a custom render.

**Expected behavior**
Only ReactiveSearch's query parameters should be cleared from the URL.

**Desktop (please complete the following information):**
 - Ubuntu 18.04
 - All
 - All

PR #1619, opened 2021-03-01
fix(web & vue): url-params-provider removes all params from the url

anik-ghosh-au7:
- Issue Type - Bug (#1615)

- Description - Currently the URLParamsProvider component removes all query parameters from the URL by clicking the Clear All button in the SelectedFilters irrespective of whether the query was set by a ReactiveSearch component or not. The fix was to add a check whether the query key matches any ReactiveSearch components Id before deleting it from the params in the URLParamsProvider component inside web and vue.

https://www.loom.com/share/9f60d3103a014c18920718ed5fb7f790

### DIFF
--- a/packages/vue/src/components/URLParamsProvider.jsx
+++ b/packages/vue/src/components/URLParamsProvider.jsx
@@ -106,7 +106,9 @@ const URLParamsProvider = {
 
 				if (!currentComponents.length) {
 					Array.from(this.params.keys()).forEach(item => {
-						this.params.delete(item);
+						if(this.searchComponents.includes(item)) {
+							this.params.delete(item);
+						}
 					});
 					this.pushToHistory();
 				}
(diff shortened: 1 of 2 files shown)


packages/vue/src/components/URLParamsProvider.jsx
```diff
@@ -106,7 +106,9 @@ const URLParamsProvider = {
 
 				if (!currentComponents.length) {
 					Array.from(this.params.keys()).forEach(item => {
-						this.params.delete(item);
+						if(this.searchComponents.includes(item)) {
```
bietkul: @anik-ghosh-au7 Add type checks before calling `includes` method.

packages/web/src/components/basic/URLParamsProvider.js
```diff
@@ -108,7 +109,9 @@ class URLParamsProvider extends Component {
 
 			if (!currentComponents.length) {
 				Array.from(this.params.keys()).forEach((item) => {
-					this.params.delete(item);
+					if (this.props.searchComponents.includes(item)) {
```
bietkul: Deconstruct is and add type check

bietkul (changes requested):
@anik-ghosh-au7

bietkul (approved):

Issue #1631, opened 2021-03-23
ReactiveList pagination is not resized correctly when size prop is changed


niroshanjwm:
**Affected Projects**
React Native 

**Library Version:** 
0.12.0

**Describe the bug**
ReactiveList pagination is not correctly resized when the size prop changes.

**To Reproduce**
- Implement ReactiveSearch with ReactiveList.
- Change the size props to a different value (In my case I implement a select-box to change the size)
- Check the pagination count according to changed size and total records, Can see pagination is not resized according to the size prop and total count.

**Expected behavior**
After the size prop changed pagination should be resized according to size prop and total count

**Desktop (please complete the following information):**
 - OS: macOS

**Smartphone (please complete the following information):**
 - Device: iPhone 11
 - OS: 13.2.2.

niroshanjwm:
Created a pull request with a fix

PR #1632, opened 2021-03-23
Fix/pagination resize on size props

niroshanjwm:
This is a fix for #1631

**Code Changes**
Update the if condition to allow status update on size prop change


bietkul (approved):
lgtm 💯

PR #1634, opened 2021-03-24
add transformResponse prop to native ReactiveBase

niroshanjwm:
add `transformResponse` prop to native ReactiveBase component, which is not added previously and available in web

### DIFF
--- a/packages/native/src/components/basic/ReactiveBase.js
+++ b/packages/native/src/components/basic/ReactiveBase.js
@@ -31,6 +31,7 @@ class ReactiveBase extends Component {
 			app: props.app,
 			credentials,
 			type: this.type,
+			transformResponse: props.transformResponse
 		};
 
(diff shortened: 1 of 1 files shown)


siddharthlatest:
Original response: @NiroshanKJ I would recommend instead to use the React Native Searchbox library. We support `transformResponse` there, https://docs.appbase.io/docs/reactivesearch/react-native-searchbox/quickstart/ and intend to maintain it actively.

*Edit:* I'm not against adding this prop in the RN ReactiveSearch, but we aren't actively maintaining that lib.

niroshanjwm:
Hi @siddharthlatest,
I am working on a native app that heavily relies on this library, currently, we don't have any plan to change it. So it will be really helpful if you can accept this change 😊
Thanks for the support!

bietkul (approved):

PR #1639, opened 2021-03-29
Change prev & next with an arrow

quintenbuis:
Because of the text not being translatable there is an issue with translations. This will eliminate the need to translate the text and makes the buttons internationally known.


bietkul (changes requested):
@quintenbuis Thanks for the PR but we can not release this in the next patch version because it would break for existing users. You can make those texts customizable by defining new slots `prev-page-label` and `next-page-label`.

PR #1646, opened 2021-04-01
Button instead of anchor

royduin:
As anchors do have/support a `disabled` attribute and they don't have a `href` so it doesn't make much sense to use anchors over buttons.

### DIFF
--- a/packages/vue/src/styles/Button.js
+++ b/packages/vue/src/styles/Button.js
@@ -6,7 +6,7 @@ const filters = ({ colors: { borderColor } }) => css`
 	margin: 0 -3px;
 	max-width: 100%;
 
-	a {
+	button {
 		margin: 2px 3px;
 		padding: 5px 8px;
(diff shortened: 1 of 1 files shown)


bietkul (changes requested):
@royduin Can you please share a demo for this? Would it not be a breaking change for existing users?

royduin:
I'm a little busy but I'm coming back on this soon!

royduin:
image

Nothing changes and everything just works as with anchors but with real support for the `disabled` attribute. Running into issues with anchors as the CSS selector `a:disabled` doesn't work as anchors do not have a `disabled` attribute. `button:disabled` works beacuse buttons really support `disabled`. And why use an anchor when there is no `href`?

Is it a breaking change? If anyone styles the pagination by element like `a { ... }` it's breaking, otherwise not.

bietkul (approved):

Issue #1736, opened 2021-07-07
ToggleButton should not unselect active item - There should always be an item active


larsdouweschuitema:
**Affected Projects**
VueJS

**Is your feature request related to a problem? Please describe.**
ToggleButton unselects value when clicked on active item

**Describe the solution you'd like**
ToggleButton should not unselect value when clicked on active item. It should only change the active item when clicked on that active item

**Describe alternatives you've considered**
Not considered. But could be a custom reactive component.

**Additional context**
Could be a prop to have different behavior in case you want to continue to offer this option in both ways

**Steps to reproduce**
1. Initial state:
![image](https://user-images.githubusercontent.com/8817968/124781296-932dfe80-df43-11eb-9779-1692dab69aa7.png)

2. Change active item:
![image](https://user-images.githubusercontent.com/8817968/124781324-99bc7600-df43-11eb-9c24-31482aa46f7c.png)

3. Click active item (same item as step 2) - Undesired behavior
![image](https://user-images.githubusercontent.com/8817968/124781368-a345de00-df43-11eb-9548-a5a17b75432b.png)

4. Click active item (same item as step 2) - Desired behavior / Stays the same
![image](https://user-images.githubusercontent.com/8817968/124781324-99bc7600-df43-11eb-9c24-31482aa46f7c.png)

bietkul:
@mohdashraf010897 We can define a new prop named `enableStrictSelection` similar to `SingleList`.
We would need that change in `React` and `Vue` libraries.
![alt img](https://i.imgur.com/RUnSj2D.png)

PR #1737, opened 2021-07-08
feat(web/ vue): added enable_strict_selection prop to toggle-button

mohdashraf010897:
`Enhancement` 

**Description:** 
ToggleButton should not unselect value when clicked on an active item. It should only change the active item when clicked on that active item.
For solving this purpose, we have introduced the `enableStrictSelection` prop which solves the purpose when used with `multiSelect = false`.

**Libraries Affected** :
- RS: WEB
- RS: VUE

[Loom Demo](https://www.loom.com/share/11f1e4730054475fbaaab59e42f359c4)

[Docs PR](https://github.com/appbaseio/Docs/pull/211)

**Feature Request PR:** #1736

### DIFF
--- a/packages/vue/examples/toggle-button/src/App.vue
+++ b/packages/vue/examples/toggle-button/src/App.vue
@@ -15,6 +15,8 @@
 							{ label: 'Adventure', value: 'Adventure' },
 							{ label: 'Music', value: 'Music' },
 						]"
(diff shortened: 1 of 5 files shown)


bietkul (approved):
@mohdashraf010897 lgtm 👍

Issue #1590, opened 2021-01-19
Catch 401 errors without throwing error when auth is not valid anymore


rap2hpoutre:
When a search fails and returns 401 (let's say because the session is not valid anymore), Reactive Search throws an error. I would love to catch it to _not trigger an error_, since it just means the connection is lost. The problem in throwing an error is that it adds some mess the console and the bug trackers (eg: Sentry). For instance, I would love a clean way to catch the error then redirect the user to a auth. `onError` on components is too late since the error is already thrown.

I'm using v3.14.0 (React project)

bietkul:
It seems a good idea. We can add `onError` prop in the `ReactiveBase` to catch errors.

DanielHabenicht:
Love it and was about to create another Issue for it!
I also think it would be a great Idea to make this for all Errors that might happen, not just the network requests.

rap2hpoutre:
Any news about this issue?

siddharthlatest:
Thanks for bringing attention to this. We would like to support this use-case. Also in case someone wants to contribute sooner, happy to provide guidance on how to do this.

rap2hpoutre:
@siddharthlatest OK, thank you. I don't know where to start, but I would be glad to tackle this issue. I checked ReactiveBase in web but I did not found where the query is actually done nor where I can catch errors (I see a `componentDidCatch` that does not seem to be triggered). Could you give me guidance on how I should start?

bietkul:
@rap2hpoutre The existing `componentDidCatch` would not get invoked on network calls error. The best way to catch the network requests errors is to use the [StateProvider](https://docs.appbase.io/docs/reactivesearch/v3/advanced/stateprovider/) component.
Check this example https://codesandbox.io/s/optimistic-euclid-2tbm7?file=/src/index.js

bietkul:
@rap2hpoutre As per your use-case, you want to handle the error before `ReactiveSearch` processes it. For e.g if `ReactiveList` was showing some results and next request caused some error then it should not clear the results.
- To do this we can define a global error handler in `ReactiveBase`.
```js
<ReactiveBase
   onError={(error) => new Promise((resolve, reject) => {
       // resolve would forward the error to RS
       if (error.code === 401) {
          // avoid processing and redirect to login modal
           reject(error)
       } else {
          // forward the error to RS
          resolve(error)
       }
   })}
/>
```
- We can store the prop in Redux store like `transformRequest`.
[...]

PR #1799, opened 2021-10-10
feat(web): add global onError

rap2hpoutre:
Add global `onError` to handle (for instance) 401 errors without throwing error when auth is not valid anymore.

See: https://github.com/appbaseio/reactivesearch/issues/1590
Implementation details: https://github.com/appbaseio/reactivesearch/issues/1590#issuecomment-938752096

Not sure it's correctly implemented, let me know if I can improve it.


mohdashraf010897:
@rap2hpoutre The prop definition is fine but the passing and storing `onError` in the redux store isn't sufficient to leverage the intended functionality.

The problem is that we are handling query processing and error handling in the `reactivecore` library, which is where the onError is supposed to be used in addition to passing it in ReactiveBase. 
I doubt you missed the third bulleted point mentioned by @bietkul @ https://github.com/appbaseio/reactivesearch/issues/1590#issuecomment-938752096.

PR #1806, opened 2021-11-02
Feat/rs searchbox

mohdashraf010897:
**PR Type** For ReactiveSearch, we will create a Searchbox component in ReactiveSearch instead of modifying the existing DataSearch component itself

- [x]  Searchbox support in React
- [x]  Searchbox support in Vue https://github.com/appbaseio/reactivesearch/pull/1842

[🎥  Loom React](https://www.loom.com/share/e8d804bd14924b57b4060eec5af03c99)

[Notion Link](https://www.notion.so/appbase/8648760a66da4dcfa5ebcefd2e439ddc)

[Docs PR Link ](https://github.com/appbaseio/Docs/pull/245)

**Related PR(s)**
- https://github.com/appbaseio/playground/pull/72
- https://github.com/appbaseio/reactivecore/pull/110


packages/web/src/components/shared/AutofillSvg.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import styled from '@emotion/styled';
+
+const AutofillSvgIcon = styled.button`
```
bietkul, line 4: @mohdashraf010897 We already have an autofill icon that is being used in `DataSearch`. Is there any reason to define a new component?

bietkul (changes requested):
# Functionality issues
- There are three queries at initial load, but you said that there would be two. Can you clarify?
alt de
- What is wrong with category suggestions? Why so many duplicates?

- Highlighting is off for category suggestions. Check how do we display category suggestions in `CategorySearch`.
- How do you handle category selection?
- How can be `_key` empty in aggs?
[](https://i.imgur.com/NZ6tT2Q.png)
- We use material icons. Promoted icon doesn't look good. It should match the icon family of recent and popular suggestions.
[](https://i.imgur.com/Y5FWStD.png)
- Throw error if `enableAppbase` is not set to `true` and `SearchBox` component is used.

packages/web/src/components/search/SearchBox.js
```diff
@@ -0,0 +1,1039 @@
...
+	parseFocusShortcuts,
+	suggestionTypes,
+} from '../../utils';
+import Mic from './addons/Mic';
+import CancelSvg from '../shared/CancelSvg';
+import CustomSvg from '../shared/CustomSvg';
+import SuggestionWrapper from './addons/SuggestionWrapper';
+import AutofillSvg from '../shared/AutofillSvg';
+
+class SearchBox extends Component {
```
bietkul: @mohdashraf010897 Can we avoid using class components?

bietkul (changes requested):

packages/web/src/components/search/SearchBox.js
```diff
@@ -0,0 +1,1230 @@
...
+			triggerCustomQuery(valueParam);
+			triggerDefaultQuery(valueParam);
+		} else if (props.autosuggest) {
+			triggerDefaultQuery(valueParam);
+		} else {
+			triggerCustomQuery(valueParam);
+		}
+	}, props.debounce);
+
+	function setValue(
```
bietkul: @mohdashraf010897 Can you use the same pattern (using `const`) to define the functions?

packages/web/src/components/search/SearchBox.js
```diff
@@ -0,0 +1,1230 @@
...
+					customQuery: true,
+					value,
+					isOpen,
+				}),
+			);
+		}
+		onValueSelected('', causes.CLEAR_VALUE, null);
+	};
+
+	function shouldMicRender(showVoiceSearch) {
```
bietkul: here too

bietkul (changes requested):

packages/web/src/components/search/SearchBox.js
mohdashraf010897: @bietkul fixed here https://github.com/appbaseio/reactivesearch/pull/1806/commits/0a8aff3dfb6c6cf9c72ff92b1c72c7d037d2c3ea

packages/web/src/components/search/SearchBox.js
mohdashraf010897: @bietkul fixed here 0a8aff3

PR #1820, opened 2021-12-01
fix: proposed fix for customQuery issue

mohdashraf010897:
**PR Type** `BugFix`

**Description** 
- `CustomQuery` and `DefaultQuery` props have a function signature where we provide the component `value` as the first argument. It helps user to change the query based on the current value. For e.g

```jsx
customQuery = (value, props) {
			return {
					query: {
							match: {title: ${value}}
					}
			}
}
```

-  The above functionality is not working with `ReactiveComponent` when `URLParams` is enabled.

[📔 Notion](https://www.notion.so/appbase/Custom-Query-issue-in-ReactiveSearch-96c8bfad62444d1c9bfe367df471874a)

[Loom Demo](https://www.loom.com/share/cf03f0c93bd540c9a828dc5f0d33dc4d)

### DIFF
--- a/packages/vue/src/server/index.js
+++ b/packages/vue/src/server/index.js
@@ -38,19 +38,19 @@ function getValue(state, id, defaultValue) {
 			const parsedValue = JSON.parse(state[id]);
 			return {
 				value: parsedValue,
-				reference: 'URL'
+				reference: 'URL',
 			};
 		} catch (error) {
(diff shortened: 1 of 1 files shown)


bietkul:
The issue has been fixed here https://github.com/appbaseio/reactivesearch/commit/7ff9b7c4f5f42f89f8d0a8bc3866d14105a3fdf6

PR #1856, opened 2022-01-19
fix(web): defaultSuggestions bug with DataSearch & CategorySearch

mohdashraf010897:
**PR Type** 🐞 `Fix`

**Description** The PR fixes `defaultSuggestions` bug with the DataSearch & CategorySearch components.

[📔 Notion](https://www.notion.so/appbase/ReactiveSearch-Issues-930ee67a8bbc44eaad66c72398ccb437#59e5756a4c4d41dd9a2292fc13aedad7l)

https://www.loom.com/share/7dd6195f1b55466297eb420f910ac24f
https://www.loom.com/share/521abbbc22f44d348f41e065fa1d1b9f

**Storybook**
- DataSearch https://opensource.appbase.io/playground/?path=/story/search-components-datasearch--with-defaultsuggestions
- CategorySearch https://opensource.appbase.io/playground/?path=/story/search-components-categorysearch--with-defaultsuggestions


packages/web/src/components/search/DataSearch.js
```diff
@@ -1059,9 +1059,8 @@ class DataSearch extends Component {
 		const {
 			theme, themePreset, size, recentSearchesIcon, popularSearchesIcon,
 		} = this.props;
-		const hasSuggestions = currentValue
-			? suggestionsList.length || this.topSuggestions.length
-			: this.defaultSuggestions.length;
+		const hasSuggestions
```
bietkul: @mohdashraf010897 Instead of changing the condition here, we should update the getter method for `defaultSuggestions`. The current change would create the side effects because `suggestionList` and `topSuggestions` can have stale values.

bietkul (changes requested):

bietkul (approved):

PR #1853, opened 2022-01-15
chore(vue): Update vue-slider-component

mohdashraf010897:
**PR Type** `Bug Fix`

**Description** 
- https://github.com/appbaseio/reactivesearch/issues/1815
   - updated `vue-slider-component` to **v3.2.15**.
   - Demo 👉 https://www.loom.com/share/867f7300455348b6b24c9a9344628868
   - Demo With SSR 👉 https://www.loom.com/share/01c3650665134332840bd460979580b9

[📔  Notion](https://www.notion.so/appbase/ReactiveSearch-issues-86750916b182477793caefc4d89b377f)

PR #1886, opened 2022-02-19
Round dynamic range slider values

royduin:
The slider was updated recently: https://github.com/appbaseio/reactivesearch/pull/1853 and [by default the interval is 1](https://nightcatsama.github.io/vue-slider-component/#/api/props?hash=interval) which doesn't work with decimals. You'll get a console error:
```
The prop "interval" is invalid, "(max - min)" must be divisible by "interval"
```
And the slider doesn't work as it should. This rounds the values to fix this issue.

> Internal reference: RAP-255

### DIFF
--- a/packages/vue/src/components/range/DynamicRangeSlider.jsx
+++ b/packages/vue/src/components/range/DynamicRangeSlider.jsx
@@ -326,11 +326,11 @@ const DynamicRangeSlider = {
 						<vue-slider-component
 							ref="slider"
 							value={[
-								Math.max(start, this.currentValue[0]),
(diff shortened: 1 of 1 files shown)


royduin:
Took some time to figure it out how to use this already but found it:
```
yarn add 'https://gitpkg.now.sh/royduin/reactivesearch/packages/vue?slider-rounding&scripts.postinstall=yarn%20install%20--ignore-scripts%20%26%26%20yarn%20run%20build-es&scripts.build-es=nps%20build.es'
```

royduin:
Thanks for merging; when will it be released?

Issue #1704, opened 2021-06-02
ReactiveComponent re renders custom component within render props


aoloo:
**Affected Projects**
React

**Library Version:** x.y.z
^3.18.1

**Describe the bug**
When utilizing `ReactiveComponent` and rendering a custom component in `render` prop. ReactiveComponent re renders numerous number of times causing performance issues. 

**To Reproduce**
https://codesandbox.io/s/reactive-component-re-renders-ynqth?file=/src/index.js
In a small code sandbox implementation, the sandbox console outputs the re rendered value within the render props.  

Within  a bare bones implementation in production environment the component is rendered more than 200 times.
**Expected behavior**
The component in render props should only be rendered once than multiple times.

**Screenshots**
![image](https://user-images.githubusercontent.com/66189547/120541891-73417300-c3b8-11eb-8e90-110f527ecda0.png)
![image](https://user-images.githubusercontent.com/66189547/120542088-aedc3d00-c3b8-11eb-9f4b-ab18fd74e025.png)

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser: [Chrome ]
 - Version [90.0.4430.212 ]

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

siddharthlatest:
@aoloo Thanks for bringing attention to this, I'm able to reproduce it. We will be fixing this in the next release.

The same issue is replicable with: 
- &lt;ReactiveGoogleMap>, &lt;ReactiveOpenStreetMap> (renders 4 - 6x) and
- &lt;ReactiveList> https://codesandbox.io/s/fervent-raman-25si4 (renders 2x)

components as well.

PR #1906, opened 2022-03-02
Remove fallback values causing re-renders

SavvyShah:
- Remove some fallback values such as `[]` which have different references each time to stop unnecessary re-renders.
- Related to #1704

### DIFF
--- a/packages/web/src/components/basic/ReactiveComponent.js
+++ b/packages/web/src/components/basic/ReactiveComponent.js
@@ -267,7 +267,7 @@ class ReactiveComponent extends Component {
 			hits, aggregations, aggregationData, promotedResults, rawData,
 		} = this.props;
 		let filteredResults = parseHits(hits);
-		if (promotedResults.length) {
+		if (promotedResults && promotedResults.length) {
 			const ids = promotedResults.map(item => item._id).filter(Boolean);
 			if (ids) {
 				filteredResults = filteredResults.filter(item => !ids.includes(item._id));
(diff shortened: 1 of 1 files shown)


bietkul:
@SavvyShah Can you share a demo to explain the fix?

packages/web/src/components/basic/ReactiveComponent.js
```diff
@@ -350,18 +350,17 @@ ReactiveComponent.propTypes = {
 ReactiveComponent.componentType = componentTypes.reactiveComponent;
 
 const mapStateToProps = (state, props) => ({
-	aggregations:
-		(state.aggregations[props.componentId] && state.aggregations[props.componentId]) || null,
-	aggregationData: state.compositeAggregations[props.componentId] || [],
-	hits: (state.hits[props.componentId] && state.hits[props.componentId].hits) || [],
+	aggregations: state.aggregations[props.componentId],
+	aggregationData: state.compositeAggregations[props.componentId],
+	hits: state.hits[props.componentId] && state.hits[props.componentId].hits,
```
bietkul, line 355: @SavvyShah Are you handling `null` value for hits? Component should not break. The default value of `hits` should be empty array.

packages/web/src/components/basic/ReactiveComponent.js
```diff
@@ -350,18 +350,17 @@ ReactiveComponent.propTypes = {
 ReactiveComponent.componentType = componentTypes.reactiveComponent;
 
 const mapStateToProps = (state, props) => ({
-	aggregations:
-		(state.aggregations[props.componentId] && state.aggregations[props.componentId]) || null,
-	aggregationData: state.compositeAggregations[props.componentId] || [],
-	hits: (state.hits[props.componentId] && state.hits[props.componentId].hits) || [],
+	aggregations: state.aggregations[props.componentId],
+	aggregationData: state.compositeAggregations[props.componentId],
```
bietkul, line 354: @SavvyShah Are you handling `null` value for `aggregationData`? The default value of `aggregationData` in `render` method and `onData` callback should be empty array.

bietkul (changes requested):

PR #1916, opened 2022-03-11
fix(web): searchbox bugs

mohdashraf010897:
**PR Type** 🐞 `Fix`

**Description** The PR fixes:
- [x] Dropdown height
- [x] Result query trigger issue with category suggestions

https://www.loom.com/share/c6d2cf97fdc54472af6e657cee255647


bietkul (changes requested):
@mohdashraf010897 Style change is a breaking change for components which can not be released in a minor release. It would be best if you can only do it for `SearchBox` component.

bietkul (changes requested):
@mohdashraf010897 
1. Share a loom to override the `suggestion-list` style using `innerClass` prop.
2. What is the need to update snapshot for dropdown list components?
3. Resolve conflicts

https://www.loom.com/share/a1e68c7174e446ffa3685544b625c670

bietkul (approved):

PR #1965, opened 2022-05-09
bug pagination

QuangHuy312:


siddharthlatest:
What is this PR for, can you share a description? You should also only include the changes without changing the lint. I'm closing it for the moment.

PR #1996, opened 2022-06-09
fix(web): remove searchbox extra section header margins

mohdashraf010897:
**PR Type** 🐞 `Fix`

**Description** The PR,

1.  removes the extra section header margins 
2. Adds a `customEvents` param to function subAction

https://www.loom.com/share/e2735904a3e444f4bfc0c64c7877f65e

### DIFF
--- a/packages/web/src/components/search/SearchBox.js
+++ b/packages/web/src/components/search/SearchBox.js
@@ -1040,7 +1040,7 @@ const SearchBox = (props) => {
 															className="section-container"
 															key={`${item[0].sectionId}`}
 														>
-															<div
+															{XSS(sectionHtml) && <div
 																className={`section-header ${getClassName(
 																	props.innerClass,
 																	'section-label',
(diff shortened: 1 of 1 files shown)


packages/web/src/components/search/SearchBox.js
```diff
@@ -1040,15 +1041,17 @@ const SearchBox = (props) => {
...
-																}}
-															/>
+															{XSS(sectionHtml) && (
+																<div
+																	className={`section-header ${getClassName(
+																		props.innerClass,
+																		'section-label',
+																	)}`}
+																	dangerouslySetInnerHTML={{
+																		__html: XSS(sectionHtml),
```
bietkul, line 1051: @mohdashraf010897 You're computing it twice, which could be an expensive task.

bietkul (changes requested):

bietkul (approved):

PR #2045, opened 2022-08-09
fix: encoded input value 

mohdashraf010897:
**PR Type** `Fix` 🐞 

**Description** The PR fixes the issue with `SearchBox` component wherein the input value had encoded strings. Additionally, for React Searchbox the max height to featuredSuggestions icon is assigned.

Libs Affected
- [x] React
- [x] Vue

https://www.loom.com/share/b359113aa3bc4f2482d2ebbb35f020e7
https://www.loom.com/share/2a71e2b350d84518bf6fa8022ed5f3f8


packages/web/src/utils/index.js
```diff
@@ -272,3 +272,9 @@ export const getValueArrayWithinLimits = (currentValueArray, rangeArray) => {
 		return currentValueArray;
 	}
 };
+
+export function decodeHtml(html) {
+	const txt = document.createElement('textarea');
```
bietkul: @mohdashraf010897 `document` would be undefined in SSR. Handle it.

packages/vue/src/utils/index.js
```diff
@@ -204,3 +204,8 @@ export const debounce = (method, delay) => {
 		method();
 	}, delay);
 };
+export function decodeHtml(html) {
```
bietkul: @mohdashraf010897 `document` would be undefined in SSR. Handle it.

packages/vue/src/components/basic/SelectedFilters.jsx
```diff
@@ -148,7 +148,7 @@ const SelectedFilters = {
 							component,
 							`${component}-${index + 1}`,
 							() => this.remove(component, value),
-							`${selectedValues[component].label}: ${valueToRender}`,
+							`${selectedValues[component].label}: ${decodeHtml(valueToRender)}`,
```
bietkul, line 151: Why do you need to use decode html in selected filters?
The selected value should be decoded.

bietkul (changes requested):

packages/vue/src/utils/index.js
siddharthlatest: Can use something like this: https://stackoverflow.com/a/7394814

```
function parseHtmlEntities(str) {
    return str.replace(/&#([0-9]{1,3});/gi, function(match, numStr) {
        var num = parseInt(numStr, 10); // read num as normal number
        return String.fromCharCode(num);
    });
}
```

packages/vue/src/components/basic/SelectedFilters.jsx
mohdashraf010897: https://www.loom.com/share/ab112bf4045a48dcb05fb63c71720890

siddharthlatest (approved):
lgtm (y)

PR #2046, opened 2022-08-09
fix(web): use label instead of value

SavvyShah:
Use label instead of value for `TabDataList` component
- [Issue](https://www.loom.com/share/7f3e1c080c18466dac381da74970b54c)
- [Solution](https://www.loom.com/share/7dba5356e4594cc0a1de8f982900ad17)

[Reactive core](https://github.com/appbaseio/reactivecore/pull/134)


bietkul (approved):
Looks good 👍

PR #2050, opened 2022-08-17
fix: icon colors to adopt theme primaryColor

mohdashraf010897:
**PR Type** `Fix`

**Description** The PR fixes the colors of svg icons to adopt the primaryColor defined in the theme/ themePreset prop in the SearchBox component.

https://www.loom.com/share/6c7a0bff0c494e89bb6bcb4707a3d43d


packages/web/src/components/search/addons/Mic.js
```diff
@@ -129,7 +129,9 @@ class Mic extends React.Component {
 		const { className } = this.props;
 		switch (status) {
 			case STATUS.active:
-				return <ListenSvg className={className} onClick={this.handleClick} />;
+				return (
```
bietkul, line 136: @mohdashraf010897 It seems like a formatting issue.

packages/web/src/components/search/addons/Mic.js
```diff
@@ -129,7 +129,9 @@ class Mic extends React.Component {
...
+					<ListenSvg className={`${className} --listen-svg`} onClick={this.handleClick} />
```
bietkul: Why is space b/w `className` and `--listen-svg`?

packages/web/src/components/shared/SearchSvg.js
```diff
@@ -1,3 +1,4 @@
+import types from '@appbaseio/reactivecore/lib/utils/types';
```
bietkul: Only import the property which is being used.

packages/web/src/styles/IconWrapper.js
```diff
@@ -9,15 +9,35 @@ const IconWrapper = styled.div`
 	cursor: pointer;
 	height: 100%;
 	min-width: 20px;
-
-	svg.search-icon {
+	svg,
+	svg g#el_D93PK3GbmJ {
```
bietkul, line 13: @mohdashraf010897 What is this dynamic value?

packages/web/src/styles/IconWrapper.js
```diff
@@ -9,15 +9,35 @@ const IconWrapper = styled.div`
...
+		#el_-Vm65Ltfy7,
+		#el_9bggsfQOtU,
+		#el_zclQ34fvf7,
+		#el_aa9sjx4H0vA,
+		#el_z5u6RAFhx7d,
+		#el__ZcqlS20zcw,
+		#el_FYYKCI_u24e,
+		#el_RMT1KUfbdF8,
+		#el_uzZNtK32Zi,
+		g#el_QJeJ_2CDw5 > path:first-child {
```
bietkul, line 35: What are these dynamic values?

bietkul (changes requested):

packages/web/src/components/search/addons/Mic.js
mohdashraf010897: because they are different classes, one is an internal class(--listen-svg) and the other one is a class passed by the user

packages/web/src/components/search/addons/Mic.js
mohdashraf010897: not a formatting issue
https://www.loom.com/share/0f0b5c9a1f494957951237e9222d0e94

packages/web/src/components/search/addons/Mic.js
bietkul: Use `getClassName` function

PR #2064, opened 2022-09-07
fix(web): fix debounce issue with controlled usage in datasearch

mohdashraf010897:
**PR Type** `Fix` 🐛 

**Description** The PR fixes an issue with `debounce` when `<DataSearch />` is used as a controlled component.

https://www.loom.com/share/d29c6f4f99c7413d9342637e37441cf5

### DIFF
--- a/packages/web/src/components/search/DataSearch.js
+++ b/packages/web/src/components/search/DataSearch.js
@@ -631,15 +631,15 @@ class DataSearch extends Component {
 		checkValueChange(props.componentId, value, props.beforeValueChange, performUpdate);
 	};
 
-	handleTextChange = debounce((value) => {
+	handleTextChange = (value) => {
 		if (this.props.autosuggest) {
 			this.updateDefaultQuery(value, this.props);
 		} else {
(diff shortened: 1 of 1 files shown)


packages/web/src/components/search/DataSearch.js
```diff
@@ -631,15 +631,15 @@ class DataSearch extends Component {
...
 		if (this.props.autosuggest) {
 			this.updateDefaultQuery(value, this.props);
 		} else {
 			this.updateQuery(value, this.props);
 		}
-	}, this.props.debounce);
+	};
 
-	updateDefaultQuery = (value, props) => {
+	updateDefaultQuery = debounce((value, props) => {
```
bietkul, line 642: @mohdashraf010897 I think should only use `debounce` wrapper when `debounce` prop is defined. The `debounce` method might be using the `setTimeout` which can create side effects when `debounce` is zero. You can also handle it in `debounce` method to return the same method if `debounce` is `zero`|`undefined`|`null`.

bietkul (changes requested):

PR #2089, opened 2022-10-05
fix(web): TreeList TypeScript types

samkio:
The typings for TreeList were incorrect resulting in the need for overrides. The title attribute was missing and icon, leafIcon, switcherIcon and renderItem were mandatory (although they are optional). The TreeList example here https://github.com/appbaseio/reactivesearch/blob/next/packages/web/examples/TreeList/src/index.js#L18-L25 confirms this.


mohdashraf010897:
Thanks 🙌🏻

Issue #2059, opened 2022-08-30
TreeList dataField array causing elasticsearch to return error on field


Kyokatarz:
**Affected Projects**
React

**Library Version:** x.y.z
v3.38.3

**Describe the bug**
TreeList causes Elasticsearch field term to return 400 because it doesn't allow an array.
Elasticsearch version: v7.16.1 

^ Could elasticsearch version be the problem?

**To Reproduce**
Passing `dataField` props to `TreeList` using elasticsearch v7.16.1

**Expected behavior**
It should work like other components

**Screenshots**
![Screen Shot 2022-08-30 at 12 38 20](https://user-images.githubusercontent.com/51933605/187431160-8e011cfc-9bbb-43fd-a465-e94795498ba3.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

siddharthlatest:
@Kyokatarz can you share the field mapping being used for dataField, and the network request payload?

Kyokatarz:
I'm not sure how to extract field mapping, but here is network request payload :D

```
{"preference":"treeListSensor"}
{"query":{"match_all":{}},"size":0,"aggs":{"class.keyword":{"terms":{"field":["class.keyword"],"order":{"_count":"desc"}}}}}
```

markusguenther:
I have the same issue. It works when I use the field as a string (with an error in the console that I should use an array for the data Field). We use open search instead of elastic search.

```
<TreeList
            title="Spielphasen"
            componentId="gamePhases-filter"
            dataField="gamePhasesDetails"
            showCount={true}
            showSwitcherIcon={true}
            renderItem={(label, count, isSelected) => {
                const gamePhaseObject = getGamePhaseObject(label, searchFilterEntities.gamePhases);
                return (
                    <span>
                        {gamePhaseObject.label}
                    </span>
                )
            }}
        />
```

And `gamePhasesDetails` is just a UUID that is matched to the search entry.
With `dataField={['gamePhasesDetails']}` it is not working.

**Resulting in:**
```
{"error":{"root_cause":[{"type":"x_content_parse_exception","reason":"[1:82] [terms] field doesn't support values of type: START_ARRAY"}],"type":"x_content_parse_exception","reason":"[1:82] [terms] field doesn't support values of type: START_ARRAY"},"status":400}
```

**Payload:**
```
{"preference":"gamePhases-filter"}
[...]

PR #2094, opened 2022-10-05
fix(web): Handle dataField array queries for TreeList

samkio:
Pre-requisite: https://github.com/appbaseio/reactivecore/pull/142

This change ensures that the queries generated by the TreeList are correct and adhere to the fact that `dataField` is an array of strings not a single string.
I have also added some fixes for:

* nested queries
* size not being set
* sortBy not being set
* choice of `and` or `or` for selection (defaults to `or` as in MultiList)


samkio:
This should remedy - https://github.com/appbaseio/reactivesearch/issues/2059

bietkul (approved):

PR #2115, opened 2022-10-31
fix(web): range type linechart

SavvyShah:
- Fix issue with line chart when type is `range`
[Notion card](https://www.notion.so/reactivesearch/Range-Selector-for-Line-Chart-5c4921e829b4475185bd51f35181d979)


bietkul (changes requested):
## Functionality
- It seems that line chart is broken (showing bars) when range is selected.
alt img

SavvyShah:
It is line chart but the lines are so close that it appears like a bar chart. You can view loom from 2:11.

PR #2196, opened 2023-03-10
fix(web): Support focus for autosuggest = false

mohdashraf010897:
**PR Type** `fix`

**Description** The PR fixes an issue with SearchBox where focus-shortcut would not focus the searchbox where `autosuggest = false`.

https://www.loom.com/share/fb676e934553436183b3c63ec9c7ea3c


siddharthlatest (approved):
lgtm (y)

PR #2206, opened 2023-04-08
fix(web): custom query fired on clearing input via keyboard

mohdashraf010897:
**PR Type** `fix`

**Description** The PR fixes an issue with the SearchBox component wherein the results query is fired even on clearing the input via keyboard typing action. 
- Additionally, fixes another issue with `tags` mode wherein the clicking on the clear icon of the input box cleared off the tags as well although expectation is to clear the input value only and fire a suggestions query instead.

[📔 Notion](https://www.notion.so/reactivesearch/SearchBox-fires-an-query-on-empty-value-without-a-trigger-from-user-83b0e2b46d84447a93ce7bb672c87156)

https://www.loom.com/share/5cf75fbe793848ac97e0e147088dbe0e

### DIFF
--- a/packages/web/src/components/search/SearchBox.js
+++ b/packages/web/src/components/search/SearchBox.js
@@ -548,7 +548,14 @@ const SearchBox = (props) => {
 	};
 
 	const clearValue = () => {
-		setValue('', false, props, causes.CLEAR_VALUE, true, false);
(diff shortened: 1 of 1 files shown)


siddharthlatest (approved):
lgtm (y)

PR #2207, opened 2023-04-16
fix(web): searchbox cursor shifts to end in controlled component

SavvyShah:
When SearchBox is used as a controlled component, the cursor shifts to the end while typing in between words.

This happens because:
`props.value` is the directly synced value with the input when used as a controlled component. 

[Notion Card](https://www.notion.so/reactivesearch/Search-box-controlled-usage-typing-amid-input-issue-cad3e73b60244f43addee656a77823ca)

Loom preview attached in notion comments

### DIFF
--- a/packages/web/src/components/search/SearchBox.js
+++ b/packages/web/src/components/search/SearchBox.js
@@ -1427,6 +1427,8 @@ const SearchBox = (props) => {
 											{...getInputProps({
 												className: getClassName(props.innerClass, 'input'),
 												placeholder: props.placeholder,
+												// When props.value is defined,
+												// it means SearchBox is used as a controlled component
 												value: Object.hasOwn(props, 'value') ? props.value : currentValue || '',
 												onChange: onInputChange,
 												onBlur: withTriggerQuery(props.onBlur),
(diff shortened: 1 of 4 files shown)


packages/web/src/utils/index.js
```diff
@@ -50,20 +50,6 @@ export const isIdentical = (a, b) => {
 };
 export const getValidPropsKeys = (props = {}) =>
 	Object.keys(props).filter(i => validProps.includes(i));
-/**
```
SavvyShah, line 53: This is removed since no longer needed and not referenced by any other file.

packages/web/src/utils/index.js
```diff
@@ -50,20 +50,6 @@ export const isIdentical = (a, b) => {
...
- * Handles the caret position for input components
- * @param {HTMLInputElement} e
- */
-export const handleCaretPosition = (e) => {
```
SavvyShah, line 57: This is no longer referenced in any other file.

packages/web/src/components/search/SearchBox.js
```diff
@@ -506,9 +518,6 @@ const SearchBox = (props) => {
 		if (value === undefined) {
 			setValue(inputValue, false, props, undefined, true, false);
 		} else if (onChange) {
-			// handle caret position in controlled components
-			handleCaretPosition(e);
```
SavvyShah, line 510: This didn't have any effect and the cursor still shifted to the end. I suspect that when react version is `v17` or lower then only this works.

mohdashraf010897 (approved):
LGTM

PR #2218, opened 2023-05-08
Feat/vue type declaration

SavvyShah:
Added types for vue library to have auto completion and type support in IDE.

[Notion card](https://www.notion.so/reactivesearch/Funda-Vue3-Create-TS-declaration-file-for-Reactivesearch-Vue-3-2cca41d9e65e4da68c2619a1e06fb1a4)
Loom attached in notion comments.


packages/vue/examples/with-nuxt-typescript/.npmrc
mohdashraf010897, line 1: @SavvyShah Is this required?

packages/vue/src/components/ReactiveBase/ReactiveBase.d.ts
```diff
@@ -0,0 +1,33 @@
+import * as types from '../../types.ts';
+
+export interface ReactiveBaseProps {
+	app: string;
```
mohdashraf010897: @SavvyShah I think the `app` prop shouldn't be required as using the `endpoint` prop makes it optional.

packages/vue/package.json
```diff
@@ -85,6 +86,7 @@
 		"rollup-plugin-replace": "^2.2.0",
 		"rollup-plugin-terser": "^5.1.1",
 		"rollup-plugin-vue": "^6.0.0",
-		"vue-eslint-parser": "^3.2.2"
+		"vue-eslint-parser": "^3.2.2",
+		"typescript": "^2.8.1"
```
mohdashraf010897: @SavvyShah The latest version of the typescript package is `5.0.4`, why are we using `2.x`? 

ref: https://www.npmjs.com/package/typescript?activeTab=versions

packages/vue/examples/with-nuxt-typescript/components/search.vue
```diff
@@ -0,0 +1,127 @@
+<template>
+  <div class="container">
+    <!-- eslint-disable -->
```
mohdashraf010897, line 3: @SavvyShah Can we avoid disabling eslint?

packages/vue/examples/with-nuxt-typescript/pages/index.vue
```diff
@@ -0,0 +1,19 @@
+<script setup>
+import { getServerState } from '@appbaseio/reactivesearch-vue';
+import Search from '../components/search.vue';
+
+// eslint-disable-next-line no-undef
+const route = useRoute();
+// eslint-disable-next-line no-unused-vars,no-undef
```
mohdashraf010897, line 7: @SavvyShah can we avoid disabling eslint rule? Let know if there's a reason to not do so.

packages/vue/src/components/basic/ReactiveComponentPrivate.d.ts
mohdashraf010897, line 1: @SavvyShah Shouldn't we support type for `componentType`?

image

packages/vue/src/components/list/MultiDropdownList.d.ts
mohdashraf010897, line 1: @SavvyShah Certain props aren't part of the types file. A way to track and verify is suggested here - https://www.notion.so/reactivesearch/Funda-Vue3-Create-TS-declaration-file-for-Reactivesearch-Vue-3-2cca41d9e65e4da68c2619a1e06fb1a4

The above applies to all the components.

mohdashraf010897 (changes requested):
@SavvyShah Please check the comments and make the changes/ revert to resolve the comments.

Share a loom wherever relevant.

mohdashraf010897:
@SavvyShah 

Changes requested -> 

https://www.loom.com/share/a378ccf3ca444d54a9fe8589b7c150e8

packages/vue/package.json
SavvyShah: Because the web version uses the 2.x version. So it is easier to avoid conflicts in the monorepo. I can change though there are no conflicts.

packages/vue/src/components/ReactiveBase/ReactiveBase.d.ts
SavvyShah: Done.

packages/vue/examples/with-nuxt-typescript/pages/index.vue
SavvyShah: This can't be removed. useRoute is defined by nuxt compiler and hence would be undefined here. Also this is copied from the already existing nuxt example which had this rules disabled and converted into typescript.

packages/vue/examples/with-nuxt-typescript/components/search.vue
SavvyShah: Done

packages/vue/src/components/list/MultiDropdownList.d.ts
SavvyShah: Incorrect. Look at the [common props imported in every file](https://github.com/appbaseio/reactivesearch/blob/8a5421d91bab121670009ae6eb252c8077a52839/packages/vue/src/index.d.ts). Also this declaration files are created from react as a base and hence would contain an exhaustive list of all the props. Unnecessary props are removed.

packages/vue/src/components/basic/ReactiveComponentPrivate.d.ts
SavvyShah: Done. Here I've hardcoded the values. I don't know of a better solution.

mohdashraf010897 (approved):

mohdashraf010897 (approved):

PR #2223, opened 2023-06-08
fix(vue): Remove console.error

royduin:
As it not something bad

### DIFF
--- a/packages/vue/src/components/URLParamsProvider.jsx
+++ b/packages/vue/src/components/URLParamsProvider.jsx
@@ -61,7 +61,6 @@ const URLParamsProvider = {
 					);
 				} catch (e) {
 					// Do not set value if JSON parsing fails.
-					console.error(e);
 				}
 			});
 		};


packages/vue/src/components/URLParamsProvider.jsx
```diff
@@ -61,7 +61,6 @@ const URLParamsProvider = {
 					);
 				} catch (e) {
 					// Do not set value if JSON parsing fails.
-					console.error(e);
```
mohdashraf010897, line 64: @royduin We better change to `console.warn(e);`

mohdashraf010897 (changes requested):

packages/vue/src/components/URLParamsProvider.jsx
royduin: I don't think we need a message at all just like in the ReactiveBase. If it fails it's due a non reactivesearch param which should be catched and ignored.

PR #2244, opened 2023-08-29
fix: ssr usage issue

SavvyShah:
Fix usage issue with SSR using vue-slider-component.

[Notion card](https://www.notion.so/reactivesearch/Check-on-SSR-example-with-Vue-for-correctness-224220105f654f73b8bfb0cdef09e8f9)


mohdashraf010897 (approved):
👍🏻 LGTM

---

with-ssr example is breaking and might be(very unlikely though) something to do with the PR.

PR #2254, opened 2023-10-17
[Vue] Use lazy & onChange for DynamicRangeSlider

Jade-GG:
Currently, the Vue DynamicRangeSlider will not update the search query if you click on the rail or if you use the keyboard controls.

By listening to the `change` event from the VueSlider and setting it to `lazy`, we get the same behavior as the listening `drag-end` event except clicking on the rail actually works properly.

I will add that it might be worth disabling the keyboard controls entirely with `useKeyboard={false}`. The keyboard controls will cause many requests if you, for example, hold a direction, and the lazy property does not fix that. Plus the keyboard controls are kinda bad anyway 😅


mohdashraf010897 (approved):
LGTM 👍🏻

PR #2259, opened 2023-10-23
Feat/fallback render item

SavvyShah:


packages/reactivecore
siddharthlatest, line 1: @SavvyShah You're referencing a commit that's not pushed, this should resolve: https://github.com/appbaseio/reactivecore/commit/1dd511

siddharthlatest (changes requested):
Other conflicts are resolvable by merging `next`'s changes into this branch. The submodule change in `reactivecore` can't be resolved as it's not pushed.

siddharthlatest (approved):

PR #2278, opened 2024-01-03
fix(tag mode): strict selection behavior to only allow picking from suggestions

siddharthlatest:
This PR fixes a behavior for tag mode where a user setting an input value that is not part of suggestions is not allowed when `strictSelection` mode is set.

Components this applies to:
* DataSearch [Demo](https://codesandbox.io/p/sandbox/mystifying-sinoussi-2yswsg?file=%2Fsrc%2FApp.vue%3A14%2C32)
* SearchBox [Demo](https://codesandbox.io/p/sandbox/youthful-lalande-9gq6ht?file=%2Fsrc%2FApp.vue%3A12%2C26)

For code review, the main changes are in the first two commits: [DataSearch code change](https://github.com/appbaseio/reactivesearch/pull/2278/commits/dff09a4bdbe9c7cdbae26c6dae833c6684ceb109), [SearchBox code change](https://github.com/appbaseio/reactivesearch/pull/2278/commits/874fdff15ddd3d5ff8a9d4657c553e20e9ace589)

This change is available in `reactivesearch-vue` `v1.35.3`. The changes in this PR should be cherry-picked to Vue v3 version and React v4 versions.


packages/vue/src/components/search/DataSearch.jsx
```diff
@@ -574,7 +598,7 @@ const DataSearch = {
 					if (props.strictSelection && props.autosuggest) {
 						if (cause === causes.SUGGESTION_SELECT || props.value !== undefined) {
 							this.updateQueryHandler(props.componentId, queryHandlerValue, props);
-						} else if (this.currentValue !== '') {
+						} else if (this.currentValue !== '' && !props.strictSelection) {
 							this.setValue('', true);
 						}
 					} else {
```
mohdashraf010897, line 620: :598 already checks for `strictSelection` prop which makes the check for `strictSelection` at L:608 unreachable

packages/vue/src/components/search/DataSearch.jsx
```diff
@@ -785,6 +811,10 @@ const DataSearch = {
 			this.triggerClickAnalytics(suggestion._click_id);
 			if (value === undefined) {
 				this.setValue(suggestion.value, true, this.$props, causes.SUGGESTION_SELECT);
+				this.isSuggestionSelected = true;
+				setTimeout(() => {
+					this.isSuggestionSelected = false;
+				}, 50);
```
mohdashraf010897, line 828: @siddharthlatest Can you explain the reason for async toggle of `isSuggestionSelected`?

siddharthlatest: This fixes a side effect I am observing with two calls being made to this.setValue otherwise. See the use of the variable for the second place where the second call is being made from.

packages/vue/src/components/search/DataSearch.jsx
siddharthlatest: I'm not seeing a check that would make it unreachable. At L598, the check is for isTagsMode

